PR #11, opened 2024-08-17
enables gzip accept_encoding gzip for pull operation

tank-bohr:
required for https://github.com/yegor256/judges/issues/84

Not providing a test here because:

- test with webmock is impossible as it patches typhous on a higher level than this parameter starts working (libcurl)
- test with a real server would require a more sophisticated server than currently used
- such a test would test the [capability of the typhous library](https://github.com/typhoeus/typhoeus/tree/master?tab=readme-ov-file#compression) per se

### DIFF
--- a/lib/baza-rb.rb
+++ b/lib/baza-rb.rb
@@ -103,8 +103,9 @@ class BazaRb
             home.append('pull').append("#{id}.fb").to_s,
             method: :get,
             headers: headers.merge(
-              'Accept' => 'application/octet-stream'
+              'Accept' => 'application/zip, application/factbase'
             ),
+            accept_encoding: 'gzip',
             connecttimeout: @timeout,
             timeout: @timeout
           )


tank-bohr:
@h1alexbel @davvd Please take a look, guys

0crat:
@tank-bohr Hello team member. We noticed that your branch name "`pull_unzip`" doesn't follow our naming convention. As per our [policy](https://www.yegor256.com/2014/04/15/github-guidelines.html), this results in a 12-point deduction. In the future, please name your branches after the ticket number you're working on (e.g., "`10`"). Your current balance is [+120](https://zerocracy.github.io/judges-action/zerocracy-vitals.html). Let's ensure we stick to the guidelines moving forward to maintain consistency across our projects.

tank-bohr:
close in favor of #13

Issue #122, opened 2025-08-15
PUT doesn't retry on 499 failure


yegor256:
Found it [here](https://github.com/zerocracy/judges-action/actions/runs/16983410435/job/48147788118):

```
E: Invalid response code #499 at PUT https://api.zerocracy.com/durables/458
Server seems to be in trouble, will sleep for 2 (attempt no.1)...
Failed in 1m40s
error: Invalid response code #499 at PUT https://api.zerocracy.com/durables/458
bundler: failed to load command: judges (/usr/local/bundle/bin/judges)
E: Command 'upload' failed in 1m40s
/usr/local/bundle/gems/baza.rb-0.10.1/lib/baza-rb.rb:732:in 'BazaRb#checked': Invalid response code #499 at PUT https://api.zerocracy.com/durables/458 (BazaRb::ServerFailure)
	from /usr/local/bundle/gems/baza.rb-0.10.1/lib/baza-rb.rb:896:in 'block (3 levels) in BazaRb#upload'
	from /usr/local/bundle/gems/retries-0.0.5/lib/retries.rb:46:in 'Kernel#with_retries'
	from /usr/local/bundle/gems/baza.rb-0.10.1/lib/baza-rb.rb:646:in 'BazaRb#retry_it'
	from /usr/local/bundle/gems/baza.rb-0.10.1/lib/baza-rb.rb:895:in 'block (2 levels) in BazaRb#upload'
	from <internal:kernel>:168:in 'Kernel#loop'
```

We are supposed to retry a few times before failing. However, here we failed right after the first problem.

yegor256:
again here: https://github.com/zerocracy/judges-action/actions/runs/19206402099/job/54902214516

bibonix:
Hi! I would like to work on this issue. The trace shows that `PUT` aborted on the first `499` response — `retry_if_server_failed` only retries when `ret.code >= 500`, so a `499` falls through to `checked` and is raised as `BazaRb::ServerFailure`, which `retry_it` does not rescue. I plan to (1) add a failing WebMock test that returns `499` then `200` for a `PUT` and asserts `upload` succeeds after a retry, then (2) fix `retry_if_server_failed` so it also retries on `499`. Will open a PR shortly.

bibonix:
[...]

PR #171, opened 2025-10-17
feat(#122): retry on anything

l3r8yJ:
@yegor256 take a look, please 

## Summary by CodeRabbit

* **Refactor**
  * Upload retry flow updated to add a dedicated path for handling connection-closed (499) responses alongside rate-limit (429) and server errors, and to adjust the retry wrapping order and control flow during uploads.
  * Backoff behavior preserved for consistent retries.

* **Tests**
  * Added a test verifying an upload retries once after an initial connection-closed (499) response and succeeds on retry.

### DIFF
--- a/lib/baza-rb.rb
+++ b/lib/baza-rb.rb
@@ -676,7 +676,7 @@ class BazaRb
     attempt = 0
     loop do
       ret = yield
-      if ret.code == 429 || ret.code >= 500 && attempt < @retries
+      if (ret.code == 429 || ret.code >= 500) && attempt < @retries
         attempt += 1
         seconds = 2 * @pause * attempt
         @loog.info("Server seems to be in trouble, will sleep for #{seconds} (attempt no.#{attempt})...")
--- a/lib/baza-rb.rb
+++ b/lib/baza-rb.rb
@@ -676,9 +676,9 @@ class BazaRb
     attempt = 0
     loop do
       ret = yield
-      if ret.code >= 500 && attempt < @retries
+      if ret.code.between?(429, 500) && attempt < @retries
         attempt += 1
-        seconds = @pause * (2**attempt)
+        seconds = 2 * @pause * attempt
         @loog.info("Server seems to be in trouble, will sleep for #{seconds} (attempt no.#{attempt})...")
         sleep(seconds)
         next
--- a/lib/baza-rb.rb
+++ b/lib/baza-rb.rb
@@ -676,7 +676,7 @@ class BazaRb
     attempt = 0
     loop do
       ret = yield
-      if ret.code.between?(429, 500) && attempt < @retries
+      if ret.code.between?(429, 599) && attempt < @retries
         attempt += 1
         seconds = 2 * @pause * attempt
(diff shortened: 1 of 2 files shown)


lib/baza-rb.rb
```diff
@@ -676,9 +676,9 @@ def retry_if_server_failed(&)
     attempt = 0
     loop do
       ret = yield
-      if ret.code >= 500 && attempt < @retries
+      if ret.code.between?(429, 599) && attempt < @retries
         attempt += 1
-        seconds = @pause * (2**attempt)
+        seconds = 2 * @pause * attempt
```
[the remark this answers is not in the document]
yegor256: @l3r8yJ that's a good recommendation

l3r8yJ:
@yegor256 Is PDD configured for this project? I see a small issue and want to create a new puzzle

l3r8yJ:
@yegor256 fixed, take a look, please

lib/baza-rb.rb
```diff
@@ -668,6 +674,25 @@ def retry_if_server_busy(&)
...
 
+  # Execute a block with retries on 499 status codes.
+  #
+  # @yield The block to execute with retries
+  # @return [Object] The result of the block execution
+  def retry_if_connection_closed(&)
+    attempt = 0
+    loop do
+      ret = yield
+      if ret.code == 499 && attempt < @retries
```
yegor256: @l3r8yJ we never return such codes from the server

lib/baza-rb.rb
```diff
@@ -649,6 +649,12 @@ def retry_it(&)
     with_retries(max_tries: @retries, rescue: TimedOut, &)
   end
 
+  # @todo #122:30min Replace all `retry_with_**` methods.
```
edmoffo, line 652: The TODO names `retry_with_**` methods but the actual methods in this file are `retry_if_**` (`retry_if_server_busy`, `retry_if_connection_closed`, `retry_if_server_failed`). The wildcard prefix in the TODO will not match anything when someone greps for the methods to replace.

edmoffo (commented):
The new `retry_if_connection_closed` wrapper does what its name says and the test covers the 499 path. One typo in the TODO comment on `lib/baza-rb.rb` flagged inline. CI is green and the change is in scope.

kreinba (commented):
Leaving this as a comment-only review. The CI is green and the change does close zerocracy/baza.rb#122 for the PUT path, but two unresolved inline threads from prior reviewers still flag real issues that should be settled before merge: the stale assertion message on `test/test_baza-rb.rb` line 757 that claims "due to 499, 500 retries" when the stub now returns only 499, and the TODO on `lib/baza-rb.rb` line 652 whose `retry_with_**` wildcard does not match the actual `retry_if_**` method prefix used in the file. The branch is also currently in a CONFLICTING state against master and will need a rebase. No new inline comments from me.

edmoffo:
@kreinba Same here — nothing more to add until #171 rebases clean off `master` and CI re-runs against the updated head.

PR #182, opened 2025-11-01
fix #157, Add  X-Zerocracy-Host listening

### LINKED ISSUE #157: the `X-Zerocracy-Host` header must be taken into account
When we do `upload()` and `download()` in the `BazaRb`, we should listen to the `X-Zerocracy-Host` header returned back to us by the server. The value in the header is the hostname of the server where we should send our next request. The original name provided in the `@host` must be discarded and new name must be used.

This is necessary for multi-server setup, when we start uploading to one server and don't want to continue to another server. We need to stay with the same server all the time.

paveltrty:
Added handling of the `X-Zerocracy-Host` header in `upload()` and `download()` methods:
- After each server response, check for the `X-Zerocracy-Host` header
- If the header is present and contains a new hostname, update `@host` and URI
- All subsequent requests within the same operation use the new host

## Summary by CodeRabbit

* **New Features**
  * Added automatic host switching for download and upload operations. The system can now dynamically redirect to alternate endpoints when instructed by the server, improving reliability and operational flexibility.

### DIFF
--- a/lib/baza-rb.rb
+++ b/lib/baza-rb.rb
@@ -828,6 +828,12 @@ class BazaRb
           ("ranged as #{ret.headers['Content-Range'].inspect}" if ret.headers['Content-Range'])
         ]
         ret = checked(ret, [200, 206, 204, 302])
+        sticky = ret.headers && ret.headers['X-Zerocracy-Host']
+        if sticky && !sticky.empty? && sticky != @host
+          @loog.debug("Switching host from #{@host} to #{sticky} as per X-Zerocracy-Host")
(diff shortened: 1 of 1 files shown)


0crat:
@paveltrty Hi there! I noticed that the branch name "`master`" doesn't follow our [naming convention](https://www.yegor256.com/2014/04/15/github-guidelines.html). According to our team policy, this results in a **-6 point** deduction. For future branches, please use the ticket number as the branch name - in this case, "`181`" would have been perfect. Thanks for your understanding! 🚀

PR #279, opened 2026-05-19
docs: fix retries YARD default from 3 to 5 in BazaRb#initialize

SAY-5:
The YARD comment on `lib/baza-rb.rb:65` documented `retries:` as `(default: 3)`, but the keyword default on line 69 is `5`. This aligns the doc with the code. Closes #276.

### DIFF
--- a/lib/baza-rb.rb
+++ b/lib/baza-rb.rb
@@ -62,7 +62,7 @@ class BazaRb
   # @param [String] token Your Zerocracy API authentication token
   # @param [Boolean] ssl Whether to use SSL/HTTPS (default: true)
   # @param [Float] timeout Connection and request timeout in seconds (default: 30)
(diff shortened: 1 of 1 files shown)


edmoffo (approved):
Verified against the keyword default on the `def initialize` line: the YARD said 3, the signature uses 5, this aligns them. No other knock-on doc to update; the rest of the README and the surrounding comments do not name a retries default. Closes #276 cleanly.

kreinba (approved):
The YARD on line 65 now reads `(default: 5)`, matching the keyword default on line 69, which closes #276. No other doc or README reference names the retries default, so the one-line edit is the full scope of the fix. The FOSSA `License Compliance` check is red on the same SHA on `master`, so it is pre-existing and unrelated to a docs-only diff.

PR #283, opened 2026-05-20
#281 drop unused faraday and backtrace runtime dependencies

kreinba:
The gemspec declared five runtime dependencies that no code under `lib/` requires or references: `faraday`, `faraday-http-cache`, `faraday-multipart`, `faraday-retry`, and `backtrace`. The library does all of its HTTP work through Typhoeus, so the Faraday stack and `backtrace` were dead weight on every install. Issue #281 calls for dropping the five `s.add_dependency` lines and regenerating `Gemfile.lock`.

Installing `baza.rb` no longer pulls the Faraday gems, `faraday-net_http`, `multipart-post`, `net-http`, or `backtrace` as transitive runtime gems. The gemspec now lists only the gems the library actually uses: `elapsed`, `iri`, `loog`, `retries`, `tago`, and `typhoeus`. No source under `lib/` changed because none of the removed gems were referenced there.

Ran `grep -rn 'faraday\|Faraday\|backtrace\|Backtrace' lib/ test/` and got no matches, confirming the gems were unused. Ran `bundle install` to refresh `Gemfile.lock`; the only deletions are the five removed gems and their transitive entries. Ran `bundle exec rake rubocop` which reported no offenses, and `bundle exec rake test`; the same seven errors and one failure appear on master before the change (IPv6 socket and JSON encoding issues from the sandbox), so they are environmental and not caused by this change.

Closes #281

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,12 +2,7 @@ PATH
   remote: .
   specs:
     baza.rb (0.0.0)
-      backtrace (~> 0.4)
       elapsed (~> 0.2)
-      faraday (~> 2.13)
-      faraday-http-cache (~> 2.5)
-      faraday-multipart (~> 1.1)
-      faraday-retry (~> 2.3)
       iri (~> 0.11)
       loog (~> 0.6)
       retries (~> 0.0)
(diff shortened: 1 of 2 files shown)


bibonix (approved):
Verified locally: a fresh `grep -rn 'faraday\|Faraday\|backtrace\|Backtrace' lib/ test/ README.md baza.rb.gemspec` returns nothing after the diff lands, and the Gemfile.lock changes are limited to the five removed gems and their transitive entries. The single commit is scoped, the body documents the verification path, and the rake checks on both macOS and Ubuntu are green. The FOSSA red check predates this branch and the same minitest socket errors appear on master, so they are environmental rather than regressions from this change.

PR #284, opened 2026-05-21
docs: fix retries YARD default from 3 to 5 in BazaRb#initialize

SAY-5:
The YARD comment on `lib/baza-rb.rb:65` documented `retries:` as `(default: 3)`, but the keyword default on line 69 is `5`. This aligns the doc with the code. Closes #276.

### DIFF
--- a/lib/baza-rb.rb
+++ b/lib/baza-rb.rb
@@ -62,7 +62,7 @@ class BazaRb
   # @param [String] token Your Zerocracy API authentication token
   # @param [Boolean] ssl Whether to use SSL/HTTPS (default: true)
   # @param [Float] timeout Connection and request timeout in seconds (default: 30)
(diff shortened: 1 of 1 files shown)


kreinba (approved):
The YARD comment on line 65 now matches the keyword default on line 69, which is the exact change issue #276 asked for. No code paths shift, CI is green on the rake matrix, and the only red signal is the pre-existing FOSSA license check on master. Ready to merge.

edmoffo (approved):
Doc default now matches the keyword default on line 69. Diff is minimal and scope-bound to the YARD line on lib/baza-rb.rb:65.

bibonix (approved):
Docstring now matches the actual default of 5 for the retries keyword argument on BazaRb#initialize at lib/baza-rb.rb:62. One-line scope, no behavior change, CI green across the matrix. No inline comments.

PR #291, opened 2026-05-23
#288 reject non-Integer ids in five job-ID methods

### LINKED ISSUE #288: Job-ID methods accept non-Integer IDs while durable-ID methods reject them
In `lib/baza-rb.rb`, the five job-ID methods `pull` (143-144), `finished?` (162-163), `stdout` (179-180), `exit_code` (196-197), and `verified` (213-214) validate `id` with only `id.nil?` and `id.positive?`. The four durable-ID methods `durable_save` (340-342), `durable_load` (357-359), `durable_lock` (373-375), and `durable_unlock` (393-395) additionally require `id.is_a?(Integer)`.

Both groups document the parameter as `[Integer] id` and feed it into the same `home.append(...)` URL builder, so the type contract is identical, but a `Float` such as `42.5` passes the job-side guard (because `42.5.positive?` is `true`) and gets interpolated into the path as `pull/42.5.fb`, producing a request the server has no route for. `BazaRb::Fake#assert_id` enforces `is_a?(Integer)` for both groups uniformly, so the real client is the laxer of the two and the Fake will not catch this miscall in tests.

The fix is to add `raise The ID of the job must be an Integer unless id.is_a?(Integer)` between the nil-guard and the positive-guard on each of the five job-ID methods, matching the three-line pattern already used by the durable-ID methods.

kreinba:
The five job-ID methods `pull`, `finished?`, `stdout`, `exit_code`, and `verified` in `lib/baza-rb.rb` validate `id` with only `id.nil?` and `id.positive?`. The four durable-ID methods (`durable_save`, `durable_load`, `durable_lock`, `durable_unlock`) additionally require `id.is_a?(Integer)`. Both groups document the parameter as `[Integer] id` and feed it into the same `home.append(...)` URL builder, so the type contract is identical, but a `Float` such as `42.5` passes the job-side guard (`42.5.positive?` is `true`) and gets interpolated into the path as `pull/42.5.fb`, producing a request the server has no route for. `BazaRb::Fake#assert_id` enforces `is_a?(Integer)` for both groups uniformly, so the real client was the laxer of the two and the Fake would not catch this miscall in tests.

The change adds one line, `raise 'The ID of the job must be an Integer' unless id.is_a?(Integer)`, between the existing nil-guard and the positive-guard on each of the five methods, matching the three-line pattern the durable-ID methods already use. No control-flow changes beyond rejecting the previously-accepted `Float` and other non-Integer numeric inputs at the entry point.

Local checks on a clean clone: `bundle exec rake test` runs 79 tests; the 5 new `*_raises_when_id_is_not_integer` cases pass. The 7 `Errno::EAFNOSUPPORT` errors on `::1` come from `RandomPort` / `with_sinatra_server` cases that need IPv6 loopback and reproduce on master in the same sandbox (CI exposes IPv6 and they pass there). The single `test_durable_load_in_chunks` failure is a pre-existing UTF-8/US-ASCII encoding mismatch on master, also reproduced before this branch was created. `bundle exec rubocop lib/baza-rb.rb test/test_baza_edge.rb` is clean. `bundle exec yardoc --fail-on-warning lib/**/*.rb` is clean.

Closes #288

### DIFF
--- a/lib/baza-rb.rb
+++ b/lib/baza-rb.rb
@@ -141,6 +141,7 @@ class BazaRb
   # @raise [ServerFailure] If the job doesn't exist or pull fails
   def pull(id)
     raise 'The ID of the job is nil' if id.nil?
+    raise 'The ID of the job must be an Integer' unless id.is_a?(Integer)
     raise 'The ID of the job must be a positive integer' unless id.positive?
     data = ''
     elapsed(@loog, level: Logger::INFO) do
@@ -160,6 +161,7 @@ class BazaRb
   # @raise [ServerFailure] If the job doesn't exist
   def finished?(id)
     raise 'The ID of the job is nil' if id.nil?
+    raise 'The ID of the job must be an Integer' unless id.is_a?(Integer)
     raise 'The ID of the job must be a positive integer' unless id.positive?
     fin = false
     elapsed(@loog, level: Logger::INFO) do
@@ -177,6 +179,7 @@ class BazaRb
   # @raise [ServerFailure] If the job doesn't exist or retrieval fails
   def stdout(id)
     raise 'The ID of the job is nil' if id.nil?
+    raise 'The ID of the job must be an Integer' unless id.is_a?(Integer)
     raise 'The ID of the job must be a positive integer' unless id.positive?
     stdout = ''
     elapsed(@loog, level: Logger::INFO) do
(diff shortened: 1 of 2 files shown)


bibonix (approved):
The five new `unless id.is_a?(Integer)` guards in `pull`, `finished?`, `stdout`, `exit_code`, and `verified` are exactly the missing third line that the four durable-ID methods already carry on lines 346, 363, 379, and 399 of `lib/baza-rb.rb`, so the type contract is now uniform across all nine ID methods and a `Float` like `42.5` no longer leaks into `home.append(...)` as `pull/42.5.fb`. The five `*_raises_when_id_is_not_integer` cases in `test/test_baza_edge.rb` cover one method each with the exact error message the new guard raises, mirror the `test_*_raises_when_id_is_nil` pattern that already exists for the durable-ID side, and pass locally on a clean clone. No other ID-taking entry points on the public surface need the same guard. CI checks `rake (ubuntu-24.04, 3.3)`, `rake (macos-15, 3.3)`, `rubocop`, `yard`, `pdd`, and `xcop` are all green on `a8223d0`; the only red check is FOSSA `License Compliance` against `master`, unrelated to this diff. Closes #288 cleanly with no scope creep.

Issue #290, opened 2026-05-23
BazaRb::Fake#durable_load rejects target paths that real BazaRb accepts


kreinba:
`BazaRb::Fake#durable_load` at `lib/baza-rb/fake.rb:161-164` calls `assert_file(file)`, which at `lib/baza-rb/fake.rb:262-265` raises unless `File.exist?(file)` and `File.size(file).positive?`. The real `BazaRb#durable_load` at `lib/baza-rb.rb:356-365` only enforces `id.nil?`, `id.is_a?(Integer)`, `id.positive?`, and `file.nil?`; it then delegates to `download` at `lib/baza-rb.rb:794-797`, which runs `FileUtils.rm_f(file); FileUtils.touch(file)` before writing the response into it. The target path is allowed to be absent or empty, and the real client creates and overwrites it.

The Fake imposes a stricter invariant on the same argument than the real client and rejects calls the real client would accept. A consumer writing `baza.durable_load(42, '/tmp/out.bin')` against the real client passes; the same call against the Fake raises `RuntimeError: The file must exist`. The existing test at `test/test_fake.rb:80-89` only works around this by writing `'hello'` into the file before calling `durable_load`, which is the opposite of the contract the real client documents. The Fake is supposed to mirror the real surface so test doubles do not lie; here it lies in the strict direction, which is harder to spot than a permissive lie because every test that uses Fake must add scaffolding the production code does not need.

The smallest fix is to replace `assert_file(file)` in `durable_load` with `raise 'The \"file\" of the durable is nil' if file.nil?`, matching the guard the real `BazaRb#durable_load` already uses, and then drop the workaround write in `test/test_fake.rb:82` so the Fake test exercises the same call shape the real method accepts.

kreinba:
@yegor256 take a look when you get a moment; happy to clarify anything if the file or line references read off.

PR #292, opened 2026-05-23
#282 accept chunk_size: kwarg in BazaRb::Fake#push and #durable_save

edmoffo:
Closes #282. The class doc on `lib/baza-rb/fake.rb:11` promises that `BazaRb::Fake` implements the same public interface as `BazaRb`, but `Fake#push` and `Fake#durable_save` omitted the `chunk_size:` keyword that `BazaRb#push` (`lib/baza-rb.rb:116`) and `BazaRb#durable_save` (`lib/baza-rb.rb:339`) accept. Any production caller that passes `chunk_size:` works against the real client and crashes against the Fake with `ArgumentError: unknown keyword: :chunk_size`.

Both Fake methods now declare `chunk_size: BazaRb::DEFAULT_CHUNK_SIZE` and validate the value is a positive Integer; the bodies still ignore it, so the keyword exists only for signature parity. Two new tests in `test/test_fake.rb` (`test_push_accepts_chunk_size_kwarg`, `test_durable_save_accepts_chunk_size_kwarg`) lock the contract on both methods.

Verified locally with `bundle exec ruby -Ilib -Itest test/test_fake.rb` — 18 tests, 0 failures, 0 errors. Rubocop runs clean across the project (`12 files inspected, no offenses detected`). The unrelated network-test failures on master (encoding issues in `test/test_baza.rb`) are pre-existing and untouched by this change.

### DIFF
--- a/lib/baza-rb/fake.rb
+++ b/lib/baza-rb/fake.rb
@@ -34,11 +34,13 @@ class BazaRb::Fake
   # @param [String] name The unique name of the job on the server
   # @param [String] data The binary data to push to the server
   # @param [Array<String>] meta List of metadata strings to attach to the job
+  # @param [Integer] chunk_size Size of each chunk in bytes (accepted for signature parity with BazaRb#push)
   # @return [Integer] Always returns 42 as the fake job ID
-  def push(name, data, meta)
+  def push(name, data, meta, chunk_size: BazaRb::DEFAULT_CHUNK_SIZE)
     assert_name(name)
     raise 'The data must be non-empty' if data.empty?
     raise 'The meta must be an array' unless meta.is_a?(Array)
+    raise 'The chunk_size must be a positive Integer' unless chunk_size.is_a?(Integer) && chunk_size.positive?
     42
   end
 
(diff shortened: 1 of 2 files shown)


bibonix (commented):
The two new  kwargs on  (line 39) and  (line 156) restore signature parity with  () and  (), and the two new happy-path tests pin the kwarg as accepted. CI is green except for the unrelated FOSSA  failure on master. One inline note on the new positive-Integer guard: see the comment on .

lib/baza-rb/fake.rb
```diff
@@ -34,11 +34,13 @@ def whoami
...
   # @param [String] data The binary data to push to the server
   # @param [Array<String>] meta List of metadata strings to attach to the job
+  # @param [Integer] chunk_size Size of each chunk in bytes (accepted for signature parity with BazaRb#push)
   # @return [Integer] Always returns 42 as the fake job ID
-  def push(name, data, meta)
+  def push(name, data, meta, chunk_size: BazaRb::DEFAULT_CHUNK_SIZE)
     assert_name(name)
     raise 'The data must be non-empty' if data.empty?
     raise 'The meta must be an array' unless meta.is_a?(Array)
+    raise 'The chunk_size must be a positive Integer' unless chunk_size.is_a?(Integer) && chunk_size.positive?
```
bibonix: The real `BazaRb#push` on `lib/baza-rb.rb:116` does not validate `chunk_size` and lets a negative or non-Integer value flow into `upload` and eventually `File.seek`, so this guard makes the Fake reject inputs the real client silently forwards — the same direction of stricter-than-real divergence reported in #290 for `Fake#durable_load`. The signature-parity fix the issue asks for lands the same way without the guard; consider dropping it and the matching guard on line 158 to keep the Fake as permissive as the real client, or covering the rejection path with a test in either case so a future refactor cannot relax it by accident.

edmoffo: @bibonix Dropping the guards in 1ee351a — the real `BazaRb#push` on `lib/baza-rb.rb:116` and `BazaRb#durable_save` on `lib/baza-rb.rb:339` forward the value into `upload` and `File.seek` without inspection, so the Fake mirroring them as permissive is the right shape. The kwargs stay in the signature for parity with the linked test in `test_fake.rb` exercising the happy path on each method.

PR #294, opened 2026-05-23
#289 retry on download timeout by moving checked() inside retry_it

kreinba:
BazaRb#download wraps Typhoeus::Request#run in retry_it but calls checked() after every retry block, so a libcurl operation_timedout return code surfaces from checked() outside the retry wrapper. Since Typhoeus#run does not raise on transport-level failures, the existing retry_it was dead code and the first TimedOut reached the caller on attempt one even with @retries set high.

The fix mirrors the nesting BazaRb#upload uses: wrap checked() in retry_it and keep retry_if_server_failed / retry_if_server_busy underneath. A libcurl operation_timedout now re-raises BazaRb::TimedOut from inside retry_it and triggers the @retries-bounded retry. The on_body slice buffer is reset inside the innermost block so retries do not accumulate response chunks across attempts.

Ran `bundle exec ruby -Ilib -Itest test/test_baza_edge.rb -n /test_download/` on Ruby 3.3.6: 4 tests, 7 assertions, 0 failures, 0 errors. The full test_baza_edge.rb run shows the 1 pre-existing failure (test_durable_load_in_chunks encoding) and 7 pre-existing errors (network-isolated sinatra and tcp-server tests) that also fail on origin/master in this sandbox, and no new failures introduced by this change.

Closes #289

### DIFF
--- a/lib/baza-rb.rb
+++ b/lib/baza-rb.rb
@@ -800,37 +800,39 @@ class BazaRb
     elapsed(@loog, level: Logger::INFO) do
       loop do
         slice = ''
-        ret = nil
-        retry_if_server_busy do
-          retry_if_server_failed do
-            slice = ''
-            request = Typhoeus::Request.new(
-              uri.to_s,
-              method: :get,
-              headers: headers.merge(
-                'Accept' => '*',
-                'Accept-Encoding' => 'gzip',
-                'Range' => "bytes=#{File.size(file)}-"
-              ),
-              connecttimeout: @timeout,
-              timeout: @timeout
+        ret =
+          retry_it do
+            checked(
+              retry_if_server_failed do
+                retry_if_server_busy do
+                  slice = ''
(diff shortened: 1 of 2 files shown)


edmoffo (approved):
Moving `checked` inside `retry_it` is the correct fix — `with_retries(rescue: TimedOut)` only catches what `checked` raises on `return_code == :operation_timedout`, so the previous nesting let the first timeout escape past the retry boundary. The new layout mirrors `upload` on `lib/baza-rb.rb:903-915` exactly, and the `slice = ''` reset now inside the innermost block keeps response chunks from leaking across attempts when WebMock's `to_timeout` already populated the `on_body` buffer. Verified locally: `bundle exec ruby -Ilib -Itest test/test_baza_edge.rb -n /test_download/` runs 4 tests, 7 assertions, 0 failures.

PR #296, opened 2026-05-23
#286 raise required Ruby to 3.3 to match the CI matrix

### LINKED ISSUE #286: rake CI matrix runs only Ruby 3.3 while gemspec declares >=3.0
The gemspec at `baza.rb.gemspec:11` declares `s.required_ruby_version = '>=3.0'`, but `.github/workflows/rake.yml:19` runs `bundle exec rake` against `ruby: [3.3]` only. A user on Ruby 3.0, 3.1, or 3.2 reads the gemspec, installs the gem, and assumes the codebase has been exercised on their version. The CI signal does not back that up.

This is a contract mismatch between the declared minimum and the tested minimum. Any deprecation, syntax change, or stdlib gap between 3.0 and 3.3 can land on master without a red check, and the next reported breakage on an older Ruby will be the user's surprise rather than CI's.

Either widen the matrix on `.github/workflows/rake.yml:19` to `ruby: [3.0, 3.1, 3.2, 3.3]` so every advertised version runs on every push, or raise `s.required_ruby_version` in `baza.rb.gemspec:11` to `'>=3.3'` so the gemspec and the matrix agree.

edmoffo:
The gemspec at `baza.rb.gemspec:11` declared `s.required_ruby_version = '>=3.0'`, but `.github/workflows/rake.yml:19` only runs `bundle exec rake` against `ruby: [3.3]`. The README at `README.md:22` advertised `Ruby 3.2+` on top, so consumers, the contract, and the CI signal each named a different supported floor. Closes #286.

Raising the gemspec floor to `>=3.3` and the README to `3.3+` aligns all three references to the version the matrix actually exercises. Ruby 3.0, 3.1, and 3.2 reached end-of-life on the upstream schedule, so the new floor names the lowest version still receiving upstream maintenance and the one CI proves on every push.

Verified locally on Ruby 3.3.6: `bundle exec ruby -Ilib -Itest test/test_fake.rb` runs 16 tests, 14 assertions, 0 failures, 0 errors; `rubocop baza.rb.gemspec` is clean. The same coverage warning that prints on master is unchanged by this PR (no `lib/` code was touched).

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ Read
 [these guidelines](https://www.yegor256.com/2014/04/15/github-guidelines.html).
 Make sure your build is green before you contribute
 your pull request. You will need to have
-[Ruby](https://www.ruby-lang.org/en/) 3.2+ and
+[Ruby](https://www.ruby-lang.org/en/) 3.3+ and
 [Bundler](https://bundler.io/) installed. Then:
 
 ```bash
--- a/baza.rb.gemspec
+++ b/baza.rb.gemspec
@@ -8,7 +8,7 @@ require_relative 'lib/baza-rb/version'
 
 Gem::Specification.new do |s|
   s.required_rubygems_version = Gem::Requirement.new('>= 0') if s.respond_to? :required_rubygems_version=
-  s.required_ruby_version = '>=3.0'
+  s.required_ruby_version = '>=3.3'
   s.name = 'baza.rb'
   s.version = BazaRb::VERSION
   s.license = 'MIT'


kreinba (approved):
The diff lines up the three places that named the supported Ruby floor: gemspec from `>=3.0` to `>=3.3`, README from `3.2+` to `3.3+`, matching the only version `.github/workflows/rake.yml` actually exercises. Ruby 3.0, 3.1, and 3.2 are all upstream EOL, so the floor advertised by the gem now matches the version CI proves on every push. The FOSSA License Compliance failure on CI is unrelated to the gemspec edit and is the same failure that shows up on `master`.

PR #300, opened 2026-05-23
#287 drop unused BazaRb::BadResponse exception class

### LINKED ISSUE #287: Unused exception class BazaRb::BadResponse in lib/baza-rb.rb
`BazaRb::BadResponse` is declared at `lib/baza-rb.rb:53` next to `ServerFailure`, `TimedOut`, `ConnectionFailed`, and `BadCompression`. Unlike its four siblings, `BadResponse` is never raised, never rescued, and never referenced anywhere else in the gem — a repo-wide grep for the constant returns only the declaration line, in both `lib/` and `test/`.

A public exception class signals an externally documented failure mode. `BadResponse` is part of `BazaRb`'s public surface yet no code path inside the gem can throw it, so any consumer writing `rescue BazaRb::BadResponse` is writing a clause that can never fire. The four neighbouring classes are reachable from `checked` and `unzip`, which is the expectation the cluster sets.

The smallest fix is to delete the line at `lib/baza-rb.rb:53` and drop `BadResponse` from the public surface. If the class was reserved for a planned check in `checked` — for example a response whose body or headers fail a shape check — raise it from that branch and add a test that exercises it, instead of leaving the declaration alone.

edmoffo:
Issue #287 reports that `BazaRb::BadResponse` is declared at `lib/baza-rb.rb:53` next to `ServerFailure`, `TimedOut`, `ConnectionFailed`, and `BadCompression`, but unlike its four siblings the class is never raised, never rescued, and never referenced anywhere else in the gem. A repo-wide `grep -rn 'BadResponse' lib/ test/` returns only the declaration line, confirming the public exception cannot fire from any code path inside this gem and any consumer writing `rescue BazaRb::BadResponse` is writing an unreachable clause.

The single hunk drops the comment and the class declaration on lines 52-53, leaving the four reachable exceptions in place. No other source file changes, no tests change, and no docs reference `BadResponse`, so the public surface narrows by exactly one symbol with no follow-on edits required. If a future check inside `checked` needs the same name, the class can be reintroduced together with the `raise` site and a test that exercises it.

Local checks on a clean clone: `bundle exec rake rubocop` reports 12 files inspected, no offenses; `bundle exec rake yard` reports 6 classes, 45 methods, 98.11% documented, no warnings; `bundle exec ruby -Ilib -e \"require 'baza-rb'; require 'baza-rb/fake'\"` loads the library cleanly and confirms `BazaRb::ServerFailure`, `TimedOut`, `ConnectionFailed`, and `BadCompression` still resolve while `BazaRb::BadResponse` is gone; `bundle exec rake test` shows the same seven errors and one failure on master before this branch (the `Errno::EAFNOSUPPORT` IPv6 cases under `with_sinatra_server` and the pre-existing UTF-8/US-ASCII encoding case in `test_durable_load_in_chunks`), so they are environmental rather than regressions from this diff.

Closes #287

### DIFF
--- a/lib/baza-rb.rb
+++ b/lib/baza-rb.rb
@@ -49,9 +49,6 @@ class BazaRb
   # {#retry_it} retries it as a transient failure.
   class ConnectionFailed < TimedOut; end
 
-  # Unexpected response arrived from the server.
-  class BadResponse < StandardError; end
-
   # When the server sent incorrectly compressed data.
   class BadCompression < StandardError; end
 


bibonix (approved):
APPROVE. The single-hunk delete of `BazaRb::BadResponse` at `lib/baza-rb.rb:52-53` removes the only declaration of an exception class that no code path raises or rescues: a fresh grep across `lib/`, `test/`, the gemspec, and the README on the head ref returns no other reference. The four sibling exceptions on adjacent lines (`ServerFailure`, `TimedOut`, `ConnectionFailed`, `BazaRb::BadCompression`) stay in place. CI is green on the rake matrix, rubocop, yard, xcop, pdd, copyrights, reuse, typos, markdown-lint, actionlint, yamllint; the only red signal is the pre-existing FOSSA `License Compliance` failure that the same step shows red on `master` and on every other open pull request, so it is unrelated to this diff.

PR #301, opened 2026-05-23
#293 retry POST on 429 like GET and upload PUT

kreinba:
Closes #293.

`BazaRb#post` at `lib/baza-rb.rb:772-787` wraps `Typhoeus::Request.post` in `retry_if_server_failed` only, while `BazaRb#get` at `:747-763` and the PUT inside `BazaRb#upload` at `:875-931` wrap their requests in `retry_if_server_busy` too. A `429 Too Many Requests` on any POST therefore bypasses the exponential backoff that GET and upload PUT receive and surfaces a `BazaRb::ServerFailure` on the first occurrence. The fix wraps the POST in the same `retry_if_server_busy` block, mirroring the nesting `get` uses, so every POST-driven write — `lock`, `unlock`, `durable_place`, `durable_lock`, `durable_unlock`, `transfer`, `fee`, `enter` — retries on 429 instead of failing immediately under load.

`lock`, `unlock`, `durable_place`, `durable_lock`, `durable_unlock`, `transfer`, `fee`, and the valve POST inside `enter` now retry on 429 with the same `@retries` × `@pause` × 2^attempt backoff that GET and upload PUT use, with no behavior change on any other status code (the two `*_busy` paths are otherwise identical, distinguished only by `>= 499` vs `== 429`).

Verified with `bundle exec ruby -Ilib -Itest test/test_baza_edge.rb -n test_post_retries_on_busy_server` (1 test, 1 assertion, 0 failures, 0 errors) — the new test stubs a 429 on the first POST and a 302 on the second and asserts two HTTP calls land, exactly the shape of the existing `test_download_retries_on_busy_server` and `test_upload_retries_on_busy_server`. The same test errors against `master` (0 assertions, 1 error: `ServerFailure` on first 429), confirming the fix is what makes the assertion land. `bundle exec rubocop lib/baza-rb.rb test/test_baza_edge.rb` is clean. The full edge suite carries a few pre-existing IPv6/`random-port` errors and a UTF-8 encoding diff in `test_durable_load_in_chunks` that reproduce on `master` and are unrelated to this change.

### DIFF
--- a/lib/baza-rb.rb
+++ b/lib/baza-rb.rb
@@ -773,13 +773,15 @@ class BazaRb
     retry_it do
       checked(
         retry_if_server_failed do
-          Typhoeus::Request.post(
-            uri.to_s,
-            body: params.merge('_csrf' => csrf).sort.to_h,
-            headers:,
-            connecttimeout: @timeout,
-            timeout: @timeout
-          )
+          retry_if_server_busy do
+            Typhoeus::Request.post(
+              uri.to_s,
+              body: params.merge('_csrf' => csrf).sort.to_h,
+              headers:,
+              connecttimeout: @timeout,
+              timeout: @timeout
+            )
+          end
         end,
         allowed
       )
--- a/test/test_baza_edge.rb
+++ b/test/test_baza_edge.rb
@@ -299,6 +299,24 @@ class TestBazaRbEdge < Minitest::Test
     end
   end
 
+  def test_post_retries_on_busy_server
+    WebMock.disable_net_connect!
+    stub_request(:get, 'https://example.org:443/csrf').to_return(body: 'token')
+    attempts = 0
+    stub_request(:post, 'https://example.org:443/lock/simple')
+      .to_return do |_request|
+        attempts += 1
+        if attempts < 2
+          { status: 429, body: 'Too Many Requests' }
+        else
+          { status: 302, body: '' }
+        end
+      end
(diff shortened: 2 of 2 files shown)


bibonix (approved):
APPROVE. The diff wraps the POST inside `retry_if_server_busy` so the nesting now mirrors `BazaRb#get` at `lib/baza-rb.rb:747-763` and the chunked PUT in `BazaRb#upload` exactly — `retry_it -> checked -> retry_if_server_failed -> retry_if_server_busy -> Typhoeus::Request.post`. `retry_if_server_busy` keys on `ret.code == 429` and `retry_if_server_failed` keys on `ret.code >= 499`, so adding the inner block changes the 429 path only and leaves every other status code untouched, including the 302 redirects the `allowed = [302]` default permits.

`test_post_retries_on_busy_server` stubs a 429 on the first `POST /lock/simple` and a 302 on the second, then asserts `attempts == 2` so the assertion lands only when the retry actually fired. The test fails against `master` with `BazaRb::ServerFailure` on the first 429 (verified locally) and passes on this branch (1 test, 1 assertion, 0 failures, 0 errors). Every POST-driven write — `lock`, `unlock`, `durable_place`, `durable_lock`, `durable_unlock`, `transfer`, `fee`, and the valve POST inside `enter` — now picks up the same retry shape.

CI is green on the rake matrix and on every linter; the only red signal is the pre-existing FOSSA `License Compliance` failure that reproduces on `master` and on every other open pull request, so it is unrelated to this diff.

PR #308, opened 2026-05-24
Fix fake durable_find file validation

### LINKED ISSUE #299: BazaRb::Fake#durable_find requires file to exist on disk while BazaRb#durable_find only sends it as a query name
`BazaRb#durable_find` at `lib/baza-rb.rb:412-428` validates `pname` and `file` are non-`nil` and non-empty strings, then sends `file` as a query parameter to `/durable-find`: `home.append('durable-find').add(file:, pname:)`. The file is never opened on the client; the server looks up the durable by that name.

`BazaRb::Fake#durable_find` at `lib/baza-rb/fake.rb:131-135` calls `assert_file(file)`, which at `lib/baza-rb/fake.rb:262-265` raises unless `File.exist?(file)` and `File.size(file).positive?`. The Fake therefore rejects every input the real client would accept as a plain filename, and conversely accepts any path-to-an-existing-file even when the same string sent to the real server is unlikely to be a meaningful query.

The YARD also disagrees with itself across the two classes: real names `file` as `The file name`, Fake names it as `The path to the file to upload` — and the Fake comment is wrong twice, because `durable_find` does not upload anything (it returns an existing durable id) and the value is never read as a path. Same shape as #295, #290, #288, #282: the Fake is the stricter side, the real client accepts a string the Fake rejects, and the contract drift surfaces only when production code hits the real server.

The smallest fix is to drop `assert_file(file)` in `BazaRb::Fake#durable_find` and replace it with `raise 'The "file" is nil' if file.nil?` and `raise 'The "file" may not be empty' if file.empty?` to mirror the real client at `lib/baza-rb.rb:415-416`, and to rewrite the Fake's YARD comment for `file` to `The file name` to match `lib/baza-rb.rb:410`. A regression test (`test_fake_durable_find_accepts_nonexistent_file_name`) pinning the loosened guard belongs next to the existing Fake tests.

Will-thom:
Closes #299.

This aligns BazaRb::Fake#durable_find with the real client contract: the argument is a durable file name, not a local file path to upload.

Changes:
- stop requiring the file name to exist on disk for Fake#durable_find
- keep nil/empty validation consistent with BazaRb#durable_find
- update the Fake YARD param text
- add regression coverage for a missing local file name

Validation:
- git diff --check
- GitHub Actions: all checks passed, including rake on ubuntu-24.04 and macos-15 with Ruby 3.3.

### DIFF
--- a/test/test_fake.rb
+++ b/test/test_fake.rb
@@ -88,6 +88,16 @@ class TestFake < Minitest::Test
     end
   end
 
+  def test_durable_find_accepts_file_name
+    baza = BazaRb::Fake.new
+    assert_equal(42, baza.durable_find('test-job', 'missing.txt'))
+  end
+
+  def test_durable_find_rejects_empty_file_name
+    baza = BazaRb::Fake.new
+    assert_raises(StandardError) { baza.durable_find('test-job', '') }
+  end
+
   def test_transfer
     baza = BazaRb::Fake.new
     receipt_id = baza.transfer('recipient', 1.0, 'test-payment')
--- a/lib/baza-rb/fake.rb
+++ b/lib/baza-rb/fake.rb
@@ -126,11 +126,12 @@ class BazaRb::Fake
   # Find a single durable.
   #
   # @param [String] pname The name of the job on the server
-  # @param [String] file The path to the file to upload
+  # @param [String] file The file name
   # @return [Integer] Always returns 42 as the fake durable ID
   def durable_find(pname, file)
     assert_name(pname)
-    assert_file(file)
+    raise 'The "file" is nil' if file.nil?
+    raise 'The "file" may not be empty' if file.empty?
     42
   end
 


test/test_fake.rb
```diff
@@ -88,6 +88,16 @@ def test_durable_operations
     end
   end
 
+  def test_durable_find_accepts_file_name
+    baza = BazaRb::Fake.new
+    assert_equal(42, baza.durable_find('test-job', 'missing.txt'))
+  end
+
+  def test_durable_find_rejects_empty_file_name
+    baza = BazaRb::Fake.new
```
edmoffo, line 97: The new Fake now raises with the specific message `The "file" may not be empty`, but the assertion catches the broader `StandardError` and does not pin the message. The sibling PRs #304 and #307 use `assert_raises(RuntimeError) { ... }` together with `assert_equal(...)` on the message, so the regression catches a future rewording of the guard or a swap to a different exception type. Tightening this to the same shape — and adding a sibling case for `nil` to cover the other branch on the same change — would match the local test convention.

edmoffo (commented):
Diff matches the smallest-fix in #299: `assert_file(file)` is replaced by `nil`/empty checks against the same `"The \"file\" is nil"` / `"The \"file\" may not be empty"` strings the real client raises at `lib/baza-rb.rb:415-416`, and the YARD now reads `The file name` like `lib/baza-rb.rb:410`. `bundle exec ruby -Ilib -Itest test/test_fake.rb` is green on the branch with the two new tests, and the only red CI gate is the unrelated FOSSA one. One nit inline on the test assertion strength relative to the new specific messages.

0crat:
@will-thom Hey! I noticed your branch name `fix-baza-durable-find-fake` doesn't follow our [naming convention](https://www.yegor256.com/2014/04/15/github-guidelines.html) - according to our policy, this means **-6** points deducted from your score. For next time, just use the ticket number as the branch name, so something like `307` would be perfect here! Your current score is [-6](https://zerocracy.github.io/judges-action/zerocracy-vitals.html) - you can check everything in your [Zerocracy account](https://www.zerocracy.com/) too.

GHX5T-SOL (approved):
Approved current head `7c136d22e1a982ac7d23160b80eeb8a9dfaca9ba`.

The diff matches #299: `BazaRb::Fake#durable_find` now treats `file` as a filename string like the real `BazaRb#durable_find`, while preserving the same nil/empty guard messages. The YARD text also matches the real client contract. I agree exact message assertions would be stronger in the test, so I checked those messages directly during validation.

Validation I ran locally with `BUNDLE_PATH=/home/kali/money/.bundle/baza-rb-308`:

- `bundle exec ruby -Itest -e 'ARGV << "--no-cov"; require "./test/test_fake"; ARGV.delete("--no-cov")' -- --name '/test_durable_find_accepts_file_name|test_durable_find_rejects_empty_file_name/'` -> 2 tests, 2 assertions, 0 failures.
- `bundle exec ruby -Itest -e 'ARGV << "--no-cov"; require "./test/test_fake"; ARGV.delete("--no-cov")' -- --quiet` -> 18 tests, 16 assertions, 0 failures.
- Direct parity probe: missing filename returns `42`; nil raises `The "file" is nil`; empty string raises `The "file" may not be empty`.
- `bundle exec ruby -Itest -e 'ARGV << "--no-cov"; Dir["test/test_*.rb"].sort.each { |f| require "./#{f}" }; ARGV.delete("--no-cov")' -- --exclude /live/ --quiet` -> 69 tests, 108 assertions, 0 failures.
- `bundle exec rubocop lib/baza-rb/fake.rb test/test_fake.rb --format simple` -> 2 files inspected, no offenses.
- `ruby -c lib/baza-rb/fake.rb` and `ruby -c test/test_fake.rb` -> Syntax OK.
- `git diff --check upstream/master...HEAD` -> clean.
- `git diff --no-ext-diff upstream/master...HEAD | gitleaks stdin --no-banner --redact --exit-code 1` -> no leaks found.

Hosted Ubuntu/macOS rake checks are green. The remaining License Compliance status is external to this source/test diff and this PR does not touch dependency or license files.

PR #311, opened 2026-05-24
Add input guards to enter

Will-thom:
Closes #298.

This aligns BazaRb#enter with the validation behavior expected by the fake client and sibling public methods. Invalid arguments now fail before opening the elapsed block, issuing HTTP calls, or executing the caller-provided computation block.

Changes:
- reject nil and empty pname, badge, and why values
- keep job optional, but reject non-Integer and non-positive values when provided
- add edge-case tests for each guard
- assert invalid inputs do not execute the enter block

Validation:
- git diff --check
- GitHub Actions: all checks passed, including rake on ubuntu-24.04 and macos-15 with Ruby 3.3.

### DIFF
--- a/lib/baza-rb.rb
+++ b/lib/baza-rb.rb
@@ -504,6 +504,14 @@ class BazaRb
   # @return [String] The cached result or newly computed result from the block
   # @raise [ServerFailure] If the valve operation fails
   def enter(pname, badge, why, job)
+    raise 'The "pname" is nil' if pname.nil?
+    raise 'The "pname" may not be empty' if pname.empty?
+    raise 'The "badge" is nil' if badge.nil?
+    raise 'The "badge" may not be empty' if badge.empty?
+    raise 'The "why" is nil' if why.nil?
+    raise 'The "why" may not be empty' if why.empty?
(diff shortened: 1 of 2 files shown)


GHX5T-SOL (approved):
Current head 624095c64538fa6bb1cade54b7c8b1d30975e6ab matches #298: invalid enter arguments now fail before elapsed/retry/yield, and the regression helper asserts the block is not executed for each rejected argument shape. I also checked the added positive-job guard against the Fake contract's positive-id requirement.

Local validation on Ruby 3.3.11:

- bundle install with BUNDLE_PATH=/private/tmp/bundle-baza-rb-311
- focused enter guard tests with coverage disabled: 8 tests, 24 assertions, 0 failures
- test/test_baza_edge.rb with coverage disabled: 32 tests, 92 assertions, 0 failures
- direct all-test require with live-named tests excluded: 75 tests, 130 assertions, 0 failures
- bundle exec rubocop lib/baza-rb.rb test/test_baza_edge.rb: 2 files inspected, no offenses
- git diff --check origin/master...HEAD: clean
- diff-only gitleaks scan: clean

Hosted rake is green on both Ubuntu and macOS. The remaining external License Compliance status is ERROR, but this PR does not touch dependency or license files.

PR #315, opened 2026-05-25
Handle malformed Content-Range without hyphen

akmhatey-ai:
`BazaRb#download` currently assumes the `range` part of a `Content-Range`
header always contains a hyphen. For a malformed response such as
`Content-Range: bytes 0/10`, `range.split('-')` leaves `e` as `nil`, and the
next line raises `NoMethodError` from `e.match?` instead of the existing
`Range is not valid (...)` validation error.

This keeps the existing validation behavior and adds only a missing-end guard:
split the range into at most two parts, raise `Range is not valid (...)` when
the end part is absent, and then run the existing numeric end check.

I checked for duplicates before opening this:

- `#312` is still open, has no comments, and is labeled `bug`.
- PR searches for `Content-Range range hyphen NoMethodError 312` returned no matching PRs.
- Open PR `#310` is for `#302` and anchors the total/range regexes; it does not handle the missing-hyphen `nil.match?` crash covered here.

Validation run locally on Windows with Ruby 3.3.11:

- New regression test fails before this patch with `NoMethodError: undefined method 'match?' for nil` at `lib/baza-rb.rb:865`.
- `bundle exec ruby -Itest -e "ARGV << '--no-cov'; require_relative 'test/test_baza_edge'; ARGV.delete('--no-cov')" -- --name test_durable_load_reports_invalid_content_range_without_hyphen`
- `bundle exec ruby -Itest -e "ARGV << '--no-cov'; require_relative 'test/test_baza_edge'; ARGV.delete('--no-cov')" -- --name test_durable_load_in_chunks`
- `bundle exec rubocop lib/baza-rb.rb test/test_baza_edge.rb`
- `ruby -c lib\baza-rb.rb`
- `ruby -c test\test_baza_edge.rb`
- `git diff --check`
- `git diff -- lib\baza-rb.rb test\test_baza_edge.rb | gitleaks stdin --no-banner --redact`

Both focused tests pass after the fix, RuboCop reports no offenses, syntax
checks pass, diff whitespace is clean, and gitleaks reports no leaks. I also
attempted the project default `bundle exec rake`; it timed out after 300s on my
local Windows setup, so I am not claiming full-suite completion from this
machine.

Closes #312

### DIFF
--- a/lib/baza-rb.rb
+++ b/lib/baza-rb.rb
@@ -861,7 +861,8 @@ class BazaRb
         _, v = ret.headers['Content-Range'].split
         range, total = v.split('/')
         raise "Total size is not valid (#{total.inspect})" unless total.match?(/^\*|[0-9]+$/)
-        _b, e = range.split('-')
+        _b, e = range.split('-', 2)
+        raise "Range is not valid (#{range.inspect})" if e.nil?
         raise "Range is not valid (#{range.inspect})" unless e.match?(/^[0-9]+$/)
         len = ret.headers['Content-Length'].to_i
         break if e.to_i == total.to_i - 1
--- a/test/test_baza_edge.rb
+++ b/test/test_baza_edge.rb
@@ -154,6 +154,18 @@ class TestBazaRbEdge < Minitest::Test
     end
   end
 
+  def test_durable_load_reports_invalid_content_range_without_hyphen
+    WebMock.disable_net_connect!
+    Dir.mktmpdir do |dir|
+      file = File.join(dir, 'loaded.txt')
+      stub_request(:get, 'https://example.org:443/durables/42')
+        .with(headers: { 'Range' => 'bytes=0-' })
+        .to_return(status: 206, body: 'x', headers: { 'Content-Range' => 'bytes 0/10' })
+      error = assert_raises(RuntimeError) { fake_baza.durable_load(42, file) }
+      assert_includes(error.message, 'Range is not valid ("0")')
+    end
+  end
+
   def test_durable_load_with_broken_compression
     WebMock.disable_net_connect!
     Dir.mktmpdir do |dir|


kreinba (commented):
The diff fixes #312 with a minimal patch: split('-', 2) keeps the empty tail for a trailing hyphen, the new nil guard catches the no-hyphen case (Content-Range: bytes 0/10), and the original numeric end check is preserved. The new regression test in test/test_baza_edge.rb pins the no-hyphen branch to the existing Range is not valid (...) message. CI is red on rake (ubuntu-24.04, 3.3) and rake (macos-15, 3.3) only because test_live_full_cycle in test/test_baza_live.rb fails against the live Zerocracy host; the failure is unrelated to this diff.

GHX5T-SOL (approved):
Reviewed current head `87bb03149153a3bc04c08cd9831b9a86b51eba1a` for #312. The parser change is narrow: `split("-", 2)` preserves the end token semantics, the explicit nil guard rejects a malformed `Content-Range: bytes 0/10` before `e.to_i`, and the existing numeric-end validation remains in place. The new regression covers the no-hyphen case, while the existing chunked download test keeps the valid `bytes 0-0/*` path covered.

Validation on my side:
- `bundle exec ruby -Itest ... --name test_durable_load_reports_invalid_content_range_without_hyphen` -> 1 test, 3 assertions, 0 failures
- `bundle exec ruby -Itest ... --name test_durable_load_in_chunks` -> 1 test, 1 assertion, 0 failures
- `bundle exec rake test -- --offline` -> 92 tests, 148 assertions, 0 failures, 8 live skips
- `bundle exec rubocop lib/baza-rb.rb test/test_baza_edge.rb --format simple` -> no offenses
- `git diff --check origin/master...HEAD` -> passed
- `git diff origin/master...HEAD | gitleaks stdin --no-banner --redact --exit-code 1` -> no leaks

GitHub checks are also green on this head, including both rake jobs and License Compliance. No live probing was performed for this review.

GHX5T-SOL (approved):
Approved current head `87bb03149153a3bc04c08cd9831b9a86b51eba1a`.

The change is narrow and fixes the malformed `Content-Range` path without changing valid range handling: `range.split('-', 2)` preserves the original end parsing, and the added `e.nil?` guard turns `Content-Range: bytes 0/10` into the existing `Range is not valid (...)` error family instead of a `NoMethodError`. The regression in `test/test_baza_edge.rb` pins that no-hyphen branch through `durable_load`, so it exercises the public download path rather than only the parser line.

Validation on Kali Ruby 3.3.8:
- `bundle exec ruby -w -Ilib:test $(bundle show rake)/lib/rake/rake_test_loader.rb test/test_baza.rb test/test_baza_edge.rb test/test_fake.rb` -> 84 tests, 148 assertions, 0 failures, 0 errors, 0 skips; coverage 98.85%.
- `bundle exec rubocop lib/baza-rb.rb test/test_baza_edge.rb --format simple` -> 2 files inspected, no offenses.
- `git diff --check origin/master...HEAD` -> passed.
- `git diff origin/master...HEAD | gitleaks stdin --no-banner --redact --exit-code 1` -> no leaks found.
- GitHub readback shows all exposed checks successful, including both Ubuntu and macOS rake jobs.

I did not find a blocker.

PR #318, opened 2026-05-25
#316: guard fake push nil data

GHX5T-SOL:
Summary:
- Add the missing nil-data guard to BazaRb::Fake#push before the existing empty-data check.
- Cover the fake client behavior so nil data now raises the same documented RuntimeError message as BazaRb#push.

Closes #316

Validation:
- Reproduced before fix: bundle exec ruby -Ilib -e "require './lib/baza-rb/fake'; begin; BazaRb::Fake.new.push('test-job', nil, []); rescue => e; puts \"#{e.class}: #{e.message}\"; exit(e.is_a?(NoMethodError) ? 0 : 1); end; exit 2" -> NoMethodError: undefined method 'empty?' for nil.
- Post-fix behavior: bundle exec ruby -Ilib -e "require './lib/baza-rb/fake'; begin; BazaRb::Fake.new.push('test-job', nil, []); rescue => e; puts \"#{e.class}: #{e.message}\"; exit(e.is_a?(RuntimeError) && e.message == 'The \"data\" of the job is nil' ? 0 : 1); end; exit 2" -> RuntimeError: The "data" of the job is nil.
- bundle exec ruby -Itest -e 'ARGV << "--no-cov"; require "./test/test_fake"; ARGV.delete("--no-cov")' -> 21 tests, 20 assertions, 0 failures, 0 errors, 0 skips.
- bundle exec rubocop lib/baza-rb/fake.rb test/test_fake.rb -> 2 files inspected, no offenses detected.
- git diff --cached --check -> passed.
- gitleaks protect --staged --no-banner --redact --verbose -> no leaks found.

Not run:
- Full rake suite, because test/test_baza_live.rb makes live api.zerocracy.com calls and the current ROE for this issue is local-only.

### DIFF
--- a/lib/baza-rb/fake.rb
+++ b/lib/baza-rb/fake.rb
@@ -38,6 +38,7 @@ class BazaRb::Fake
   # @return [Integer] Always returns 42 as the fake job ID
   def push(name, data, meta, chunk_size: BazaRb::DEFAULT_CHUNK_SIZE) # rubocop:disable Lint/UnusedMethodArgument
     checkname(name)
+    raise(RuntimeError, 'The "data" of the job is nil') if data.nil?
     raise(RuntimeError, 'The data must be non-empty') if data.empty?
     raise(RuntimeError, 'The meta must be an array') unless meta.is_a?(Array)
     42
--- a/test/test_fake.rb
+++ b/test/test_fake.rb
@@ -28,6 +28,13 @@ class TestFake < Minitest::Test
     assert_equal(42, BazaRb::Fake.new.push('test-job', 'test-data', []))
   end
 
+  def test_push_raises_when_data_is_nil
+    assert_equal(
+      'The "data" of the job is nil',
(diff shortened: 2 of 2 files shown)


kreinba (approved):
Guard prepends cleanly to BazaRb::Fake#push at lib/baza-rb/fake.rb:41 with the same wording the real client uses at lib/baza-rb.rb:116, restoring Fake/real parity for nil data. Regression test in test/test_fake.rb covers the new RuntimeError path. Diff is two files, six added lines, no unrelated edits. CI red on test_live_full_cycle only, which the author flagged as the live-API test outside local scope; no inline comments.

kreinba (approved):
Fake#push now mirrors BazaRb#push at lib/baza-rb.rb:110 with the same message verbatim, and the test pins that message. Scope is limited to lib/baza-rb/fake.rb and test/test_fake.rb. The red rake matrix is the unrelated test_live_full_cycle timeout against api.zerocracy.com (Job #74104 hang), not anything this diff exercises. No inline comments.

PR #319, opened 2026-05-25
#303: avoid recomputing enter block on retry

akmhatey-ai:
﻿Fixes #303.

BazaRb#enter used to wrap result lookup, the user block, and valve persistence in one outer retry. If POST /valves exhausted its own inner timeout retries and then succeeded on the outer retry, the user block ran again for the same logical enter call.

This change keeps the result lookup retryable, computes the missing value once, then retries only valve persistence. The new regression makes two timed-out POST /valves attempts force the outer retry and verifies the call still returns after-1 with the block invoked once.

Validation:
- RED before fix: test_enter_does_not_recompute_when_valve_post_outer_retry_succeeds failed with expected after-1, actual after-2.
- Focused regression: 1 tests, 3 assertions, 0 failures, 0 errors, 0 skips.
- Adjacent focused edge tests: 2 tests, 4 assertions, 0 failures, 0 errors, 0 skips.
- Adjacent Pact enter tests: 2 tests, 2 assertions, 0 failures, 0 errors, 0 skips.
- rubocop lib/baza-rb.rb test/test_baza_edge.rb: no offenses.
- ruby -c lib/baza-rb.rb; ruby -c test/test_baza_edge.rb: syntax OK.
- git diff --check; git diff --cached --check; diff/staged gitleaks: passed.

Not claimed: full test/test_baza_edge.rb is not clean in this Windows environment because unrelated tests hit a temp-file permission error and a missing /bin/bash helper.

### DIFF
--- a/lib/baza-rb.rb
+++ b/lib/baza-rb.rb
@@ -509,12 +509,12 @@ class BazaRb
   # @raise [ServerFailure] If the valve operation fails
   def enter(pname, badge, why, job)
     elapsed(@loog, good: "Entered valve #{badge} to #{pname}") do
+      ret = retry_it { get(home.append('result').add(badge:), [200, 204]) }
+      return ret.body if ret.code == 200
+      r = yield
+      uri = home.append('valves')
+      uri = uri.add(job:) unless job.nil?
       retry_it do
-        ret = get(home.append('result').add(badge:), [200, 204])
-        return ret.body if ret.code == 200
-        r = yield
-        uri = home.append('valves')
-        uri = uri.add(job:) unless job.nil?
         post(
           uri,
           {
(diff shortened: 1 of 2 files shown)


kreinba (approved):
Lifts the yield out of the retried region at lib/baza-rb.rb:512-525, isolating GET in its own retry_it and POST in another, so the user block runs at most once per enter call as the YARD contract at lib/baza-rb.rb:497-503 implies. Regression test at test/test_baza_edge.rb:342-361 exercises the outer retry: with retries=2, the WebMock stub forces post()'s inner retry_it to exhaust on the first two timeouts, the outer retry_it then retries once more, the third POST returns 302, and calls stays at 1. Diff stays scoped to lib/baza-rb.rb and test/test_baza_edge.rb, six lines net. CI red only on test_live_full_cycle, the same live-API failure unrelated to this diff. No inline comments.

PR #320, opened 2026-05-25
#305: anchor Content-Range total guard

GHX5T-SOL:
/claim #305
Fixes #305

Summary:
- anchors the Content-Range total-size validation as `\A(?:\*|[0-9]+)\z`, so malformed totals like `*malformed` and `abc123` no longer pass just because they start with `*` or end with digits
- adds an offline WebMock regression that verifies a malformed ranged response raises `Total size is not valid` before the download loop can advance to the next range request

Validation:
- Docker Ruby 3.3: `bundle exec ruby -Itest -e 'ARGV << "--no-cov"; require "./test/test_baza_edge"; ARGV.delete("--no-cov")'` -> 37 tests, 105 assertions, 0 failures\n- Docker Ruby 3.3: `bundle exec rake test -- --offline` -> 91 tests, 145 assertions, 0 failures, 8 skips (live tests skipped via `--offline`)\n- Docker Ruby 3.3: `bundle exec rubocop` -> 12 files inspected, no offenses\n- `git diff --cached --check`\n- `gitleaks protect --staged --no-banner --redact --verbose`\n\nNo secrets, credentials, wallet material, live service calls, or production mutations were used. The broader test run used the repository offline flag to avoid live Zerocracy API calls.

### DIFF
--- a/lib/baza-rb.rb
+++ b/lib/baza-rb.rb
@@ -860,7 +860,7 @@ class BazaRb
         break if ret.code == 200
         _, v = ret.headers['Content-Range'].split
         range, total = v.split('/')
-        raise "Total size is not valid (#{total.inspect})" unless total.match?(/^\*|[0-9]+$/)
+        raise "Total size is not valid (#{total.inspect})" unless total.match?(/\A(?:\*|[0-9]+)\z/)
         _b, e = range.split('-')
         raise "Range is not valid (#{range.inspect})" unless e.match?(/^[0-9]+$/)
         len = ret.headers['Content-Length'].to_i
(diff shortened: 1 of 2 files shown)


kreinba (approved):
Regex change at lib/baza-rb.rb:863 swaps the unanchored alternation /^\*|[0-9]+$/ for the grouped, fully anchored /\A(?:\*|[0-9]+)\z/, matching the fix the linked issue spelled out and closing the hole where '*malformed' or 'abc123' slipped through. New offline regression at test/test_baza_edge.rb:303-315 wires a 206 response with Content-Range 'bytes 0-0/*malformed' through the download path and pins the resulting message 'Total size is not valid ("*malformed")'. Scope stays at one src line and one new test; the neighbor regex on the next line that the issue called out as already correct was left untouched. CI red only on test_live_full_cycle live tests, unrelated. No inline comments.

PR #323, opened 2026-05-26
#322: align fake transfer signature

### LINKED ISSUE #322: BazaRb::Fake#transfer signature uses `*` instead of `job: nil`, silently dropping the job kwarg and skipping nil/empty guards that BazaRb#transfer enforces
`BazaRb::Fake#transfer` at `lib/baza-rb/fake.rb:199-205` declares `def transfer(recipient, amount, summary, *)`, while `BazaRb#transfer` at `lib/baza-rb.rb:440-462` declares `def transfer(recipient, amount, summary, job: nil)` and writes `body['job'] = job unless job.nil?` into the POST. The Fake's trailing `*` accepts any 4th positional argument and any unknown keyword arguments by turning them into a hash bound to the splat. The result is that `fake.transfer('alice', 1.0, 'tip', job: 42)` runs silently and returns 42 while the equivalent call against `BazaRb#transfer` would send `job=42` to the server. The job argument is invisible to tests written against the Fake.

The guards inside the body diverge as well. The Fake validates the recipient with `recipient.match?(/^[a-zA-Z0-9-]+$/)`; for `nil`, `nil.match?` returns `nil`, so the guard fires with `The recipient nil is not valid`, while the real client raises `The "recipient" is nil` from an explicit `recipient.nil?` check. The Fake has no `amount.nil?` guard, so a `nil` amount falls through to `nil.is_a?(Float)` and raises `The amount  must be a Float` (note the double space from empty interpolation), while the real client raises `The "amount" is nil` first. The Fake calls `summary.empty?` with no nil guard, so a `nil` summary raises `NoMethodError: undefined method 'empty?' for nil:NilClass`, while the real client raises `The "summary" is nil`.

This is the same Fake/real drift pattern as #298, #299, #309, and #316: one side of the pair lacks a guard or signature element the other side has, and the drift surfaces only when production code switches between clients. A test written as `assert_raises(RuntimeError) { fake.transfer('alice', 1.0, nil) }` against the Fake fails with `NoMethodError` instead of the documented `RuntimeError` the real client raises.

The smallest fix is to change the Fake signature to `def transfer(recipient, amount, summary, job: nil)`, add `assert_id(job) unless job.nil?` to mirror the type check the real path implies, and replace the four `raise` lines with the same wording the real client uses: `The "recipient" is nil`, `The "amount" is nil`, `The "amount" must be Float`, `The "amount" must be positive`, `The "summary" is nil`. A regression case in `test/test_fake.rb` next to the existing `test_transfer` should pin the new behavior for nil inputs and confirm that passing `job: 42` no longer silently disappears.

GHX5T-SOL:
Closes #322.

This aligns `BazaRb::Fake#transfer` with `BazaRb#transfer` for the reported fake/real drift:

- changes the fake transfer signature from a trailing splat to `job: nil`
- validates a provided fake `job:` with the existing fake ID guard instead of swallowing it silently
- matches the real transfer nil/type/positive guard wording for recipient, amount, and summary
- adds focused fake-client regression coverage for the new signature and guards

Validation:

- `bundle exec ruby -Itest test/test_fake.rb -- --no-cov` -> 28 tests, 32 assertions, 0 failures, 0 errors
- `bundle exec rake test -- --offline --no-cov` -> 98 tests, 156 assertions, 0 failures, 0 errors, 8 skips; live tests skipped by offline mode, no production calls
- `bundle exec rubocop lib/baza-rb/fake.rb test/test_fake.rb` -> no offenses
- `bundle exec rake rubocop` -> 12 files inspected, no offenses
- `bundle exec rake yard` -> passed, 98.11% documented
- `git diff --no-ext-diff upstream/master...HEAD | gitleaks stdin --no-banner --redact --exit-code 1` -> no leaks found

### DIFF
--- a/lib/baza-rb/fake.rb
+++ b/lib/baza-rb/fake.rb
@@ -195,12 +195,17 @@ class BazaRb::Fake
   # @param [String] recipient GitHub username of the recipient
   # @param [Float] amount The amount to transfer in ƶ (zents)
   # @param [String] summary The description/reason for the payment
+  # @param [Integer] job Optional job ID to associate with this transfer
   # @return [Integer] Always returns 42 as the fake receipt ID
-  def transfer(recipient, amount, summary, *)
+  def transfer(recipient, amount, summary, job: nil)
+    raise 'The "recipient" is nil' if recipient.nil?
     raise "The recipient #{recipient.inspect} is not valid" unless recipient.match?(/\A[a-zA-Z0-9-]+\z/)
-    raise "The amount #{amount} must be a Float" unless amount.is_a?(Float)
-    raise "The amount #{amount} must be positive" unless amount.positive?
+    raise 'The "amount" is nil' if amount.nil?
+    raise 'The "amount" must be Float' unless amount.is_a?(Float)
+    raise 'The "amount" must be positive' unless amount.positive?
+    raise 'The "summary" is nil' if summary.nil?
     raise "The summary #{summary.inspect} is empty" if summary.empty?
+    assert_id(job) unless job.nil?
     42
   end
 
--- a/test/test_fake.rb
+++ b/test/test_fake.rb
@@ -131,6 +131,50 @@ class TestFake < Minitest::Test
     assert_equal('The recipient "recipient\nbad value" is not valid', err.message)
   end
 
+  def test_transfer_accepts_job_kwarg
+    receipt_id = BazaRb::Fake.new.transfer('recipient', 1.0, 'test-payment', job: 42)
+    assert_equal(42, receipt_id)
+  end
+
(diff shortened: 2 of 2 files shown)


edmoffo (approved):
APPROVE. The signature, guard order, and guard wording in lib/baza-rb/fake.rb#L198-L206 match BazaRb#transfer on lib/baza-rb.rb#L440-L445 exactly, and the assert_id check on line 205 closes the silent-drop hole that #322 named. The nine new test_transfer_* cases in test/test_fake.rb pin every nil and type branch as well as the job kwarg path. CI is green on ec2b0ff. No inline comments.

PR #343, opened 2026-05-30
#276 correct retries default in BazaRb#initialize YARD doc

bibonix:
Closes #276.

The YARD `@param` line for `retries:` in `BazaRb#initialize` documents `(default: 3)`, but the keyword default in the constructor signature is `retries: 5`. The doc string is corrected to match the actual default, so callers reading the YARD documentation get the right retry budget and any tooling that surfaces defaults from the docs is no longer off by two.

The change is a one-line edit in `lib/baza-rb.rb:58`; no source code, signature, or test fixture is touched.

Verified locally with `bundle exec rake rubocop` (12 files, 0 offenses) and `bundle exec rake yard` (45 methods documented, 96.23% coverage, 0 warnings).

### DIFF
--- a/lib/baza-rb.rb
+++ b/lib/baza-rb.rb
@@ -55,7 +55,7 @@ class BazaRb
   # @param [String] token Your Zerocracy API authentication token
   # @param [Boolean] ssl Whether to use SSL/HTTPS (default: true)
   # @param [Float] timeout Connection and request timeout in seconds (default: 30)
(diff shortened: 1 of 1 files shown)


edmoffo (approved):
Verdict: APPROVE. The YARD line for retries: now reads (default: 5), matching the keyword default on the constructor signature, and no other code, signature, or fixture is touched. CI is green across all ten checks. No inline comments.

PR #351, opened 2026-06-03
#298: add nil/empty/type guards to BazaRb#enter

VasilevNStas:
Closes #298. `BazaRb#enter` was the only public method without argument validation. Added 8 guards for pname, badge, why, and job — matching the style of sibling methods and `Fake#enter`.

### DIFF
--- a/lib/baza-rb.rb
+++ b/lib/baza-rb.rb
@@ -487,6 +487,14 @@ class BazaRb
   # @return [String] The cached result or newly computed result from the block
   # @raise [ServerFailure] If the valve operation fails
   def enter(pname, badge, why, job)
+    raise(RuntimeError, 'The "pname" is nil') if pname.nil?
+    raise(RuntimeError, 'The "pname" may not be empty') if pname.empty?
+    raise(RuntimeError, 'The "badge" is nil') if badge.nil?
+    raise(RuntimeError, 'The "badge" may not be empty') if badge.empty?
+    raise(RuntimeError, 'The "why" is nil') if why.nil?
+    raise(RuntimeError, 'The "why" may not be empty') if why.empty?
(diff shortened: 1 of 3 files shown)


GHX5T-SOL (approved):
The current head matches #298 and is safe to merge from the client-contract side.

I checked that the new guards run before any HTTP call and before the user block is yielded: a no-network probe over nil/empty `pname`, `badge`, `why`, non-Integer `job`, and non-positive `job` raised locally with `yielded=false` and `requests=0` for each invalid case. The existing enter retry/yield regression tests still pass, so this does not reintroduce the earlier transient POST re-yield issue.

Local validation on `42f92668af0f821bc36cf0fce240f610765cea0f`:

- `bundle check`
- `ruby -c lib/baza-rb.rb`
- `ruby -c test/test_baza_edge.rb`
- focused enter guard tests: 8 tests, 16 assertions, 0 failures
- focused enter yield/retry tests: 2 tests, 5 assertions, 0 failures
- full `test/test_baza_edge.rb` with coverage disabled: 56 tests, 153 assertions, 0 failures
- full `test/test_baza.rb` with coverage disabled: 26 tests, 22 assertions, 0 failures
- full `test/test_fake.rb` with coverage disabled: 40 tests, 53 assertions, 0 failures
- `bundle exec rubocop`: 12 files inspected, no offenses
- `bundle exec rake picks`
- `bundle exec rake yard`: 96.23% documented

Hosted readback: all non-rake checks pass. Ubuntu `rake` passes the added enter guards and `test_live_enter_valve`, then fails only the known live `TestBazaLive#test_live_full_cycle` stuck-job case (`Job #76635 ... did not finish in 0s`); macOS rake is canceled.

PR #352, opened 2026-06-03
#347: anchor range-end regex with \A...\z instead of ^...$

### LINKED ISSUE #347: BazaRb#download range-end guard uses line-anchored `/^[0-9]+$/` while the adjacent total-size guard uses `\A...\z`
`BazaRb#download` in `lib/baza-rb.rb` at line 829 validates the range-end token parsed from `Content-Range` with `e.match?(/^[0-9]+$/)`. The total-size token on the adjacent line 813 uses `\A(?:\*|[0-9]+)\z` after the fix for #302, but the range-end guard on line 829 still relies on `^` and `$`, which in Ruby match line boundaries rather than string boundaries.

The drift means a `Content-Range` value whose range-end carries an embedded newline (for example `bytes 0-1023\nfoo/2048`) parses to `e = "1023\nfoo"` and clears the guard, because `^[0-9]+$` matches the first line of the captured value. The just-merged `\A...\z` total-size guard rejects the same shape on the next field; only the range-end side stays permissive. The downstream `e.to_i` then truncates at the digit boundary and the loop continues on a half-trusted offset.

The smallest fix is to swap `^` and `$` for `\A` and `\z` on line 829 so it reads `e.match?(/\A[0-9]+\z/)`, matching the anchoring style now used on line 813. A regression sibling to the existing `test_download_*` cases in `test/test_baza_edge.rb`, stubbing a 206 with `Content-Range: bytes 0-12\nfoo/100` and asserting the `Range is not valid` raise, pins the new anchors.

VasilevNStas:
Closes #347. Line 837 used `^[0-9]+$` (line-anchored) while the adjacent total-size guard on line 834 uses `\A...\z`. A `Content-Range: bytes 0-12\nfoo/100` would pass the weak anchor. Changed to `/\A[0-9]+\z/`.

### DIFF
--- a/lib/baza-rb.rb
+++ b/lib/baza-rb.rb
@@ -826,7 +826,7 @@ class BazaRb
         raise(RuntimeError, "Total size is not valid (#{total.inspect})") unless total.match?(/\A(?:\*|[0-9]+)\z/)
         _b, e = range.split('-', 2)
         raise(RuntimeError, "Range is not valid (#{range.inspect})") if e.nil?
-        raise(RuntimeError, "Range is not valid (#{range.inspect})") unless e.match?(/^[0-9]+$/)
+        raise(RuntimeError, "Range is not valid (#{range.inspect})") unless e.match?(/\A[0-9]+\z/)
         break if e.to_i == total.to_i - 1
         break if total == '0'
         chunk += 1
--- a/test/test_baza_edge.rb
+++ b/test/test_baza_edge.rb
@@ -404,6 +404,22 @@ class TestBazaRbEdge < Minitest::Test
     end
   end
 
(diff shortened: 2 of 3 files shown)


kreinba (approved):
Fix matches #347 exactly: one-line swap from `^[0-9]+$` to `\A[0-9]+\z` on the range-end guard so the anchoring style is identical to the adjacent total-size check on the preceding line. The added test `test_download_rejects_range_end_with_newline` reproduces the embedded-newline case from the issue and passes in the ubuntu rake log. The only CI failure on this head is `test_live_full_cycle` in `test/test_baza_live.rb`, which hits the real network and is unrelated to this diff.


---
_Generated by [Claude Code](https://claude.ai/code)_

PR #353, opened 2026-06-03
#345 drop application/zip content-type from gzipped uploads

edmoffo:
Closes #345.

`BazaRb#zipped` was overwriting `Content-Type` with `application/zip` whenever `@compress` was true, mislabeling the gzip-encoded octet-stream as a ZIP archive. The override is dropped so the `application/octet-stream` set by `upload` survives alongside `Content-Encoding: gzip`.

Requests with compression on now carry `Content-Type: application/octet-stream` and `Content-Encoding: gzip` together; requests with compression off are unchanged. The `test_push_compressed_content` assertion is updated to pin the new contract.

`bin/rubocop` is clean on the changed files. `bin/rake test` was run locally; the unit suites pass except for the network-bound cases that need IPv6 loopback in the container, which fail identically on master (`Errno::EAFNOSUPPORT` on `::1`), and one live test that depends on the production API.

### DIFF
--- a/lib/baza-rb.rb
+++ b/lib/baza-rb.rb
@@ -586,13 +586,7 @@ class BazaRb
     body = io.string
     params.merge(
       body:,
-      headers: params.fetch(:headers).merge(
-        {
-          'Content-Type' => 'application/zip',
-          'Content-Encoding' => 'gzip',
-          'Content-Length' => body.bytesize
-        }
-      )
+      headers: params.fetch(:headers).merge({ 'Content-Encoding' => 'gzip', 'Content-Length' => body.bytesize })
     )
   end
 
--- a/test/test_baza_edge.rb
+++ b/test/test_baza_edge.rb
@@ -101,7 +101,7 @@ class TestBazaRbEdge < Minitest::Test
       with_http(200, 'yes') do |baza|
         baza.push('simple', fb.export, %w[meta1 meta2 meta3])
       end
-    assert_equal('application/zip', req.content_type)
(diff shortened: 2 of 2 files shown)


GHX5T-SOL (approved):
Approved. I checked current head `7e5204f85012eae895f678c572af31f34a71cd19` against `#345`. The change keeps upload’s `Content-Type: application/octet-stream` while adding gzip-specific `Content-Encoding` and `Content-Length`, so compressed `push` no longer mislabels the gzip stream as `application/zip`.

Validation:
- `bundle check`
- `ruby -c lib/baza-rb.rb`
- `ruby -c test/test_baza_edge.rb`
- no-network WebMock probe: one stubbed `PUT /push/simple` with `Content-Type: application/octet-stream`, `Content-Encoding: gzip`, matching `Content-Length`, and gzip body round-tripping to the original Factbase export
- focused push compression tests: `2` tests / `7` assertions
- full non-live suites: `test/test_baza_edge.rb` (`45` tests / `131` assertions), `test/test_baza.rb` (`26` / `22`), and `test/test_fake.rb` (`40` / `53`)
- `bundle exec rubocop`: `12` files, no offenses
- `bundle exec rake picks`
- `bundle exec rake yard`: `96.23%` documented

I did not run full local `bundle exec rake` because it invokes live Zerocracy API tests. Hosted Ubuntu `rake` on this head passes the touched push-compression tests and fails only `TestBazaLive#test_live_full_cycle` with `Job #75432 ... did not finish in 0s`; hosted macOS `rake` was canceled.

PR #375, opened 2026-06-05
#365: Add pname format and length validation to BazaRb#push

VasilevNStas:
## Problem

`BazaRb::Fake#checkname` validates pname format (`\A[a-z0-9-]+\z`) and length (≤32), but `BazaRb#push` only checked for nil and empty. Since pname goes directly into the URL path (`home.append('push').append(pname)`), the lack of validation risked URL injection and caused Fake/Real divergence.

## Solution

Added two guards to `BazaRb#push` after the nil check, matching `Fake#checkname` semantics:

```ruby
```

### Tests added

- `test_push_raises_when_pname_is_invalid` — uppercase pname rejected
- `test_push_raises_when_pname_is_too_long` — 33-char pname rejected

## Verification
- [x] All existing tests pass (Pact pnames like `pact72` match the pattern)
- [x] 2 new edge tests pass

Closes #365.

### DIFF
--- a/lib/baza-rb.rb
+++ b/lib/baza-rb.rb
@@ -107,6 +107,8 @@ class BazaRb
   def push(pname, data, meta, chunk_size: DEFAULT_CHUNK_SIZE)
     raise(RuntimeError, 'The "name" of the job is nil') if pname.nil?
     raise(RuntimeError, 'The "name" of the job may not be empty') if pname.empty?
+    raise(RuntimeError, "The name #{pname.inspect} is not valid") unless pname.match?(/\A[a-z0-9-]+\z/)
+    raise(RuntimeError, "The name #{pname.inspect} is too long") if pname.length > 32
     raise(RuntimeError, 'The "data" of the job is nil') if data.nil?
     raise(RuntimeError, 'The "data" of the job may not be empty') if data.empty?
     raise(RuntimeError, 'The "meta" of the job is nil') if meta.nil?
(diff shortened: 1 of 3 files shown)


GHX5T-SOL (approved):
Approved. This is a narrow fix for issue #365: `BazaRb#push` now applies the same product-name format and length guards that `BazaRb::Fake#push` already reaches through `checkname`, before the name is appended into the `/push/<pname>` URL. The added edge tests use the real `BazaRb` client helper pointed at `example.org`, so they cover the production path, not only the fake.

Validation I ran on exact head `6aad60f13ec36325c0cce00c2023c8f47690de11`:
- local no-network WebMock probe -> invalid and overlong names are rejected before any HTTP request is sent; valid `valid-name` still uploads through a stub
- focused push-name metadata subset with coverage disabled -> 3 tests, 8 assertions, 0 failures
- full `test/test_baza_edge.rb` with coverage disabled -> 50 tests, 143 assertions, 0 failures

Hosted readback: all non-rake checks pass. Ubuntu `rake` passes both new push-name tests and existing `test_pushes_to_create_job`, then fails only the known live `TestBazaLive#test_live_full_cycle` stuck-job case (`Job #76597 ... did not finish in 0s`) after 124 tests / 234 assertions. The macOS `rake` job is canceled and its log was unavailable by job ID. I did not run local full `bundle exec rake` because it invokes live Zerocracy API tests, which I kept out of scope.

PR #376, opened 2026-06-05
#366: Add pname format and length validation to lock/unlock/name_exists?/recent

VasilevNStas:
## Problem

`BazaRb::Fake#checkname` validates pname format (`\A[a-z0-9-]+\z`) and length (≤32), but four BazaRb methods only checked nil and empty:

- `lock(pname, owner)` — pname goes into `/lock/<pname>`
- `unlock(pname, owner)` — pname goes into `/unlock/<pname>`
- `name_exists?(pname)` — pname goes into `/exists/<pname>`
- `recent(name)` — name goes into `/recent/<name>.txt`

All four risk URL injection and Fake/Real divergence.

## Solution

Added format and length guards to each method after the existing nil/empty checks, matching `Fake#checkname`:

```ruby
```

### Tests added (5 new edge tests)

- `test_lock_raises_when_pname_is_invalid`
- `test_lock_raises_when_pname_is_too_long`
- `test_unlock_raises_when_pname_is_invalid`
- `test_name_exists_raises_when_pname_is_invalid`
- `test_recent_raises_when_pname_is_invalid`

## Verification
- [x] All 50 edge tests pass (5 new)
- [x] All 26 unit/Pact/fake tests pass

Closes #366.

### DIFF
--- a/lib/baza-rb.rb
+++ b/lib/baza-rb.rb
@@ -225,6 +225,8 @@ class BazaRb
   def lock(pname, owner)
     raise(RuntimeError, 'The "pname" of the product is nil') if pname.nil?
     raise(RuntimeError, 'The "pname" of the product may not be empty') if pname.empty?
+    raise(RuntimeError, "The name #{pname.inspect} is not valid") unless pname.match?(/\A[a-z0-9-]+\z/)
+    raise(RuntimeError, "The name #{pname.inspect} is too long") if pname.length > 32
     raise(RuntimeError, 'The "owner" of the lock is nil') if owner.nil?
     raise(RuntimeError, 'The "owner" of the lock may not be empty') if owner.empty?
     elapsed(@loog, level: Logger::INFO) do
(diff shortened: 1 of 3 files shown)


GHX5T-SOL (approved):
Reviewed current head `706842c2966cb825c093f0064c791a823d197e5e`; this looks correct to me.

The change adds the same `/\A[a-z0-9-]+\z/` and length <= 32 guards before each touched method appends the name into the URL. The new `test/test_baza_edge.rb` cases use `fake_baza`, which is actually a real `BazaRb` pointed at `example.org`, so they exercise the production client path.

I also checked the missing overlong variants with a local no-network WebMock probe: invalid and overlong values for `lock`, `unlock`, `name_exists?`, and `recent` all raised before any HTTP request, while valid `valid-name` reached the expected stubbed `lock`, `unlock`, `exists`, and `recent` paths.

Local validation passed: `bundle check`; syntax checks for `lib/baza-rb.rb` and `test/test_baza_edge.rb`; `git diff --check upstream/master...HEAD`; focused new edge tests; full coverage-disabled `test/test_baza_edge.rb`, `test/test_baza.rb`, and `test/test_fake.rb`; `bundle exec rubocop`; `bundle exec rake picks`; and `bundle exec rake yard`.

Hosted readback: all non-rake checks pass. Ubuntu `rake` passes the new validation tests and the existing touched client tests, then fails only `TestBazaLive#test_live_full_cycle` with a Zerocracy live job not finishing; macOS `rake` was canceled. I did not run full local `bundle exec rake` because it includes live Zerocracy API tests.

VasilevNStas:
@yegor256 fixed it, please review

PR #377, opened 2026-06-05
#367: Add pname format and length validation to durable_place/durable_find

VasilevNStas:
## Problem

`BazaRb::Fake#checkname` validates pname format (`\A[a-z0-9-]+\z`) and length (≤32), but `durable_place` and `durable_find` only checked nil and empty.

## Solution

Added format and length guards to both methods. 3 new edge tests.

## Verification
- [x] All tests pass

Closes #367.

### DIFF
--- a/lib/baza-rb.rb
+++ b/lib/baza-rb.rb
@@ -297,6 +297,8 @@ class BazaRb
   def durable_place(pname, file)
     raise(RuntimeError, 'The "pname" of the durable is nil') if pname.nil?
     raise(RuntimeError, 'The "pname" of the durable may not be empty') if pname.empty?
+    raise(RuntimeError, "The name #{pname.inspect} is not valid") unless pname.match?(/\A[a-z0-9-]+\z/)
+    raise(RuntimeError, "The name #{pname.inspect} is too long") if pname.length > 32
     raise(RuntimeError, 'The "file" of the durable is nil') if file.nil?
     raise(RuntimeError, "The file '#{file}' is absent") unless File.exist?(file)
     if File.size(file) > 1024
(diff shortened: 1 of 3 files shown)


GHX5T-SOL (approved):
Reviewed current head `cd105b92d242a856e91b94f94539ea58906aa08b`; this looks correct to me.

The change adds the same `/\A[a-z0-9-]+\z/` and length <= 32 guards before `durable_place` or `durable_find` can send the product name in the request body/query. The new `test/test_baza_edge.rb` cases use `fake_baza`, which is a real `BazaRb` pointed at `example.org`, so they exercise the production client path.

I also checked the extra overlong `durable_find` case with a local no-network WebMock probe: invalid and overlong values for both `durable_place` and `durable_find` all raised before any HTTP request, while valid `valid-name` reached the expected stubbed `/durable-place` and `/durable-find` paths.

Local validation passed: `bundle check`; syntax checks for `lib/baza-rb.rb` and `test/test_baza_edge.rb`; `git diff --check upstream/master...HEAD`; focused new edge tests; full coverage-disabled `test/test_baza_edge.rb`, `test/test_baza.rb`, and `test/test_fake.rb`; `bundle exec rubocop`; `bundle exec rake picks`; and `bundle exec rake yard`.

Hosted readback: all non-rake checks pass. Ubuntu `rake` passes the new durable-name tests and existing durable client tests, then fails only `TestBazaLive#test_live_full_cycle` with a Zerocracy live job not finishing; macOS `rake` was canceled and its job log was unavailable. I did not run full local `bundle exec rake` because it includes live Zerocracy API tests.

PR #378, opened 2026-06-05
#370: Merge await and recover into single retry_on method

VasilevNStas:
## Problem

`await` (retry on 429) and `recover` (retry on >= 499) were always used together as nested `recover { await { ... } }`. The retry logic was duplicated.

## Solution

Replaced both with a single `retry_on` method that handles both conditions in one loop. All 4 call sites simplified from nested blocks to flat `retry_on { ... }`.

- `await` and `recover` methods removed
- `retry_on` added to `.rubocop.yml` AllowedNames
- Net LoC: -22

## Verification
- [x] All tests pass

Closes #370.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -28,6 +28,7 @@ Elegant/GoodMethodName:
     - durable_unlock
     - durable_find
     - format_value
+    - retry_on
     - will_respond_with
 Elegant/ClassInModule:
   Exclude:
--- a/lib/baza-rb.rb
+++ b/lib/baza-rb.rb
@@ -606,44 +606,29 @@ class BazaRb
     with_retries(max_tries: @retries, rescue: TimedOut, &)
   end
 
-  # Execute a block with retries on 429 status codes.
+  # Retry the HTTP request on rate-limiting (429) and server errors (>= 499).
+  #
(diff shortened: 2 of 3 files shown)


GHX5T-SOL (changes requested):
The refactor changes retry behavior for mixed transient statuses.

Before this PR, `recover { await { request } }` kept separate counters: one for `>= 499` responses and one for `429` responses. With `retries: 1`, a no-network WebMock sequence of `499 -> 429 -> 200` succeeds on current `master` after 3 calls, because the server-error retry starts a fresh `await` pass and that pass can still retry the rate-limit response.

On this PR head (`80b08fbb13b253c385163b61e6a2a0722c161f53`), the same sequence fails after 2 calls with `BazaRb::ServerFailure: Invalid response code #429 at GET https://example.org:443/file`, because the single counter in `retry_on` has already been consumed by the 499 before the 429 is seen.

The existing tests cover pure 429 and pure 499 paths, and those still pass, but they do not cover this mixed transient sequence. Please either preserve the old separate retry budgets (for example, keep parameterized wrappers like `retry_on(->(r) { r.code == 429 })` and `retry_on(->(r) { r.code >= 499 })`) or explicitly document and test that the intended behavior is now one shared retry budget across both status families.

PR #384, opened 2026-06-05
#355: Make rehost opt-in to prevent SSRF via X-Zerocracy-Host

VasilevNStas:
## Problem

`BazaRb#rehost` switches `@host` to any hostname from `X-Zerocracy-Host` response header, sending the auth token to arbitrary hosts (client-side SSRF).

## Solution

- Added `allow_rehost: false` keyword argument to constructor (opt-in, disabled by default)
- `rehost` returns early unless `@allow_rehost` is true
- Updated 4 existing tests that rely on rehost to pass `allow_rehost: true`
- Added `test_rehost_rejected_by_default` to verify SSRF protection

## Verification
- [x] All tests pass (46 edge tests, 132 assertions)
- [x] New test confirms rehost is blocked by default

Closes #355.


GHX5T-SOL (approved):
Approved. The current head makes `X-Zerocracy-Host` rehosting opt-in via `allow_rehost: false`, keeps the existing host-switch behavior covered when explicitly enabled, and adds a default-deny regression. I did not find a blocking issue in the source/test surface.

Validation I ran locally on the exact head `7c656653f3ba0b3df2646d35865246d741b674b5`:
- `ruby -c lib/baza-rb.rb` -> Syntax OK
- focused `test_rehost_rejected_by_default` with coverage disabled -> 1 test, 1 assertion, 0 failures
- full `test/test_baza_edge.rb` with coverage disabled -> 49 tests, 138 assertions, 0 failures

Hosted check readback: all non-rake checks pass. In the Ubuntu `rake` log, `test_rehost_rejected_by_default`, `test_live_durable_lock_unlock`, and the other live durable/auth tests pass; the only failure is the known `TestBazaLive#test_live_full_cycle` stuck-job case (`Job #76592 ... did not finish`). The macOS job passed the relevant live durable tests before cancellation. I did not run local full `bundle exec rake` because it invokes live Zerocracy API tests, which I kept out of scope.

lib/baza-rb.rb
```diff
@@ -60,7 +60,11 @@ class BadCompression < StandardError; end
   # @param [Integer] pause The factor on pause (<1 means faster, >1 means slower)
   # @param [Loog] loog The logging facility (default: Loog::NULL)
   # @param [Boolean] compress Whether to use GZIP compression for requests/responses (default: true)
-  def initialize(host, port, token, ssl: true, timeout: 30, retries: 5, pause: 1, loog: Loog::NULL, compress: true)
+  # @param [Boolean] allow_rehost Whether to follow X-Zerocracy-Host header (default: false)
+  def initialize(
+    host, port, token, ssl: true, timeout: 30, retries: 5, pause: 1, loog: Loog::NULL, compress: true,
+allow_rehost: false
```
yegor256: @VasilevNStas can we invent a better name of a single noun? see https://www.yegor256.com/2015/01/12/compound-name-is-code-smell.html

lib/baza-rb.rb
```diff
@@ -60,7 +60,11 @@ class BadCompression < StandardError; end
   # @param [Integer] pause The factor on pause (<1 means faster, >1 means slower)
   # @param [Loog] loog The logging facility (default: Loog::NULL)
   # @param [Boolean] compress Whether to use GZIP compression for requests/responses (default: true)
-  def initialize(host, port, token, ssl: true, timeout: 30, retries: 5, pause: 1, loog: Loog::NULL, compress: true)
+  # @param [Boolean] rehost Whether to follow X-Zerocracy-Host header (default: false)
+  def initialize(
+    host, port, token, ssl: true, timeout: 30, retries: 5, pause: 1, loog: Loog::NULL, compress: true,
+rehost: false
```
morphqdd: `rehost: false` is at column 0 — the keyword arg isn't indented with the rest of the signature. This is a Layout offense and just reads badly; align it with the other kwargs on the line above (or keep the whole signature on fewer lines).

lib/baza-rb.rb
```diff
@@ -71,6 +75,7 @@ def initialize(host, port, token, ssl: true, timeout: 30, retries: 5, pause: 1,
     @pause = pause
     @compress = compress
     @mutex = Mutex.new
+    @rehost = rehost
```
morphqdd: Two naming points: (1) the PR description says the flag is `allow_rehost:`, but the code uses `rehost:`/`@rehost`. (2) `@rehost` (a boolean) collides in spirit with the `rehost(ret, uri)` method, which is presumably why you had to add an `Elegant/GoodVariableName` exception in `.rubocop.yml`. Renaming to `allow_rehost:`/`@allow_rehost` reads as a boolean, matches your own description, and lets you drop the rubocop config change entirely.

.rubocop.yml
```diff
@@ -30,6 +30,9 @@ Elegant/GoodMethodName:
     - format_value
     - retry_on
     - will_respond_with
+Elegant/GoodVariableName:
+  AllowedNames:
+    - '@rehost'
```
morphqdd: This allowlist entry only exists to work around the `@rehost` name (see the naming comment). If you rename the ivar to `@allow_rehost`, this exception can be removed so the PR doesn't loosen the linter config.

lib/baza-rb.rb
```diff
@@ -63,7 +63,11 @@ class BadCompression < StandardError; end
   # @param [Integer] pause The factor on pause (<1 means faster, >1 means slower)
   # @param [Loog] loog The logging facility (default: Loog::NULL)
   # @param [Boolean] compress Whether to use GZIP compression for requests/responses (default: true)
-  def initialize(host, port, token, ssl: true, timeout: 30, retries: 5, pause: 1, loog: Loog::NULL, compress: true)
+  # @param [Boolean] sticky Whether to follow X-Zerocracy-Host header, switching to a different host (default: false)
```
Favixx, line 66: The keyword is `sticky:` here, but the PR title and description both say `allow_rehost:`. Whichever name wins, they need to agree before merge — this is a public constructor argument, and the PR description is what people will find later when they search for how to re-enable the behaviour.

The doc line also does not say why the default is `false`. As written it reads like a transport tuning knob. It should state that enabling it makes the client follow an arbitrary hostname chosen by the server and send `X-Zerocracy-Token` there, so anyone flipping it to `true` to unblock an upload understands what they are accepting. Something closer to: `@param [Boolean] sticky Follow the X-Zerocracy-Host header and switch @host to whatever the server names; off by default because the new host is not validated and receives the auth token`.

`README.md` documents none of the client's options today, so there is nothing to update there, but this is a behaviour change to a released gem and deserves a line in the release notes.

lib/baza-rb.rb
```diff
@@ -63,7 +63,11 @@ class BadCompression < StandardError; end
...
+  def initialize(
+    host, port, token, ssl: true, timeout: 30, retries: 5, pause: 1, loog: Loog::NULL, compress: true,
+    sticky: false
```
Favixx, line 69: `sticky: false` makes the whole rehost mechanism dead code for every caller that exists. Searching GitHub for `BazaRb.new`: `zerocracy/fbe` at `lib/fbe/enter.rb:34` (`BazaRb.new('api.zerocracy.com', 443, options.zerocracy_token, loog:)`) and `yegor256/judges` at `lib/judges/commands/push.rb:36`, `pull.rb:35`, `download.rb:36` and `upload.rb:38` all build the client with `ssl`, `timeout`, `retries`, `loog` and sometimes `compress`, and none of them pass `sticky:`. There are no other consumers.

That is what makes this an effective fix today and also what makes it a regression tomorrow. `upload` has a dedicated recovery branch for `Expecting chunk #N` that logs "Server lost upload state at chunk #N, restarting from chunk #M" and rewinds; that path only makes sense if consecutive chunks must reach the same backend node, which is presumably what `X-Zerocracy-Host` exists to arrange. Turning affinity off by default across `judges push` and `judges pull` is likely to convert a working chunked upload into a rewind loop bounded by `@retries`.

The realistic sequence is: this ships, uploads start rewinding, someone adds `sticky: true` to `judges` to make them work again, and the SSRF returns in full because nothing validates the host. Please either pair this with PRs against `judges` and `fbe`, or change the design so the safe configuration is also the working one — an allowlist defaulting to the zerocracy domain and enabled by default, rather than an unvalidated switch defaulting to off.

lib/baza-rb.rb
```diff
@@ -73,6 +77,7 @@ def initialize(host, port, token, ssl: true, timeout: 30, retries: 5, pause: 1,
     @retries = retries
     @pause = pause
     @compress = compress
+    @sticky = sticky
```
Favixx, line 80: This branch is 55 commits behind `master` (`git rev-list --count $(git merge-base master HEAD)..master`). The green checks on this PR ran on 2026-06-05 against a tree that no longer exists: `master` has since changed `@pause` in this same constructor to `@pause = ENV.fetch('BAZA_BACKOFF', pause).to_i` and extracted a `backoff(attempt)` helper. I confirmed the merge is still textually clean, but nothing has run the suite against the combination.

Please rebase and let CI re-run before merge, particularly since #385 and #405 are both queued against `lib/baza-rb.rb` and one of them will land first.

While rebasing, note that `@sticky` is read in `rehost` outside `@mutex`, while `@host` is read and written inside it. That is safe only as long as `@sticky` stays write-once in the constructor. Worth a short comment saying so, because if a setter is ever added, the guard on line 564 becomes a race against the very block it protects.

lib/baza-rb.rb
```diff
@@ -556,6 +561,7 @@ def csrf
   # @return [Iri] The updated URI object (or original if no valid header present)
   # @note Invalid hostnames are logged as warnings and ignored
   def rehost(ret, uri)
```
Favixx, line 563: Worth spelling out the blast radius in the method being changed, because the new early return is now the only defence and its scope is easy to misjudge. `rehost` is called from exactly two places: in `download`, immediately after `checked(...)` and before the `blanks.include?(ret.code)` branch, and in `upload`, after each successful chunk PUT.

Because the `download` call sits before the blank-response handling, even a `204` or `302` carrying `X-Zerocracy-Host` retargets the client; the loop then reissues the ranged GET — with `Range` and the full `headers` hash, including `X-Zerocracy-Token` — against the attacker's host. In `upload` the retarget happens between chunks, so the remaining bytes of the factbase being uploaded go to the new host along with the token. And because `rehost` assigns `@host` rather than only returning a modified `Iri`, every later call on that instance (`whoami`, `balance`, `finish`, `csrf`, all the `durable_*` methods) builds its URI from `home`, which reads `@host`, and follows as well.

One header on one response therefore owns that client for the rest of its lifetime. That is why gating alone is not enough: the enabled configuration is not "slightly risky", it is complete takeover of the client's traffic and credentials from a single response header.

lib/baza-rb.rb
```diff
@@ -556,6 +561,7 @@ def csrf
...
+    return uri unless @sticky
```
Favixx, line 564: This one line is the entire fix, and it gates the feature without validating the target. `hostname?` below is a pure RFC 1123 syntax check: it rejects empty strings, anything over 253 bytes, and illegal characters, and accepts everything else. With `sticky: true` the client will still switch to `evil.com`. It will also switch to `localhost`, to any resolvable internal name, and — because digits are alphanumeric and satisfy the label rule — to a bare IPv4 literal such as `169.254.169.254`, which on a cloud runner is the instance metadata endpoint. The auth token accompanies the switch in every case.

Issue #355 asks for exactly this validation; its Option A is `allowed_hosts: ['*.zerocracy.com']`. Neither of its options is implemented. Concretely, after the syntax check add something like `return uri unless @allowed_hosts.any? { |p| File.fnmatch(p, host, File::FNM_CASEFOLD) }`, with `@allowed_hosts` defaulting to the registrable suffix of the constructor's `host` argument, so a client pointed at `api.zerocracy.com` can only ever be moved within `*.zerocracy.com`. That has the useful side effect of being safe enough to leave enabled by default, which also resolves the caller problem raised on the constructor.

As it stands, `Closes #355` overstates the change. It closes the hole for callers that never opted in — today, all of them — and leaves it fully open for the first caller that opts in, which the caller analysis suggests will happen the moment chunked uploads start rewinding.

lib/baza-rb.rb
```diff
@@ -556,6 +561,7 @@ def csrf
   # @return [Iri] The updated URI object (or original if no valid header present)
   # @note Invalid hostnames are logged as warnings and ignored
```
Favixx, line 562: This `@note` was already wrong and the PR is the right moment to fix it, because the change makes silence the normal case. It claims invalid hostnames are logged as warnings; they are not. `return uri unless hostname?(sticky)` on line 567 discards them with no log at any level, and the new `return uri unless @sticky` does the same for the disabled case.

Both silences are costly. An operator whose uploads stop sticking after this release gets no clue why, which is how a security default gets reverted rather than understood. And a server attempting to redirect the client somewhere it is not permitted to go is exactly the event you want recorded. Add `@loog.warn` on both early returns — for example `@loog.warn("Ignoring X-Zerocracy-Host: #{sticky.inspect} (host stickiness is disabled)")` for the new guard, and a matching one for a syntactically rejected hostname — then rewrite the `@note` to describe what the method actually does, including the new precondition.

Doing that requires reordering slightly: read the header first, then check `@sticky`, so the warning can name the host that was refused. That ordering is strictly better than the current one anyway, since it makes the disabled path observable instead of indistinguishable from "no header was sent".

lib/baza-rb.rb
```diff
@@ -556,6 +561,7 @@ def csrf
...
   def rehost(ret, uri)
+    return uri unless @sticky
     sticky = ret.headers && ret.headers['X-Zerocracy-Host']
     return uri unless sticky
```
Favixx, line 566: Even in the intended, trusted case the switch is permanent and one-way. `@host` is overwritten and the constructor's original host is not retained anywhere, so there is no fallback: if the client is redirected to a node that subsequently goes down, every following request — including the retries inside `attempt` and `retry_on` — targets the dead node, and the only recovery is to construct a new `BazaRb`.

That is a reliability defect in its own right, and it also sharpens the security problem, because it means a single hostile response is not a transient event but a permanent one for that client instance. Keep the constructor's host as `@origin` and re-anchor to it when the current host stops responding, or at minimum when `attempt` exhausts its retries. If the allowlist suggested on line 564 is added, `@origin` is also the natural source for its default suffix.

lib/baza-rb.rb
```diff
@@ -556,6 +561,7 @@ def csrf
...
     return uri unless hostname?(sticky)
```
Favixx, line 567: Two further properties of `hostname?` worth pinning down while this method is being touched. It permits single-label names, so `localhost` passes; on a GitHub Actions runner, where `judges` executes, that also covers whatever short internal names the runner's resolver knows about. And it permits a trailing dot, which the next line strips with `chomp('.')`, so `evil.com.` and `evil.com` are the same target — worth remembering when writing the allowlist match, since a naive string comparison against an allowed list would have to run after the `chomp`, not before.

The other thing to record in a comment here is that `rehost` changes only the host: `@port` and `@ssl` are preserved. With the default `ssl: true`, the redirect goes to `https://evil.com:443` and TLS validates normally against the attacker's own certificate. TLS provides no protection whatsoever against this class of redirect; there is nothing between the response header and the token except this method. That is worth stating in the code so nobody later assumes HTTPS makes the switch safe enough to skip validation.

test/test_baza_edge.rb
```diff
@@ -695,6 +695,21 @@ def test_rehost_is_thread_safe
     )
   end
 
+  def test_rehost_rejected_by_default
```
Favixx, line 698: This test cannot fail for any reason related to the code below the new guard. `return uri unless @sticky` is the first statement in `rehost`, so with `sticky` defaulting to false the method returns before it looks at headers at all — delete every remaining line of `rehost` and this test still passes. It asserts that the flag is wired up, which is worth having, but the name `test_rehost_rejected_by_default` promises a security property that is not being tested.

The stubbed round-trip is also doing no work. `stub_request(...)` plus a real `Typhoeus::Request.get` exists only to obtain an object that responds to `headers`. `Typhoeus::Response.new(code: 200, headers: { 'X-Zerocracy-Host' => 'evil.com' })` produces the same input with no WebMock dependency and none of the `connecttimeout`/`timeout` noise, and makes it immediately obvious to the next reader that the HTTP layer is irrelevant to the assertion.

test/test_baza_edge.rb
```diff
@@ -695,6 +695,21 @@ def test_rehost_is_thread_safe
...
+    WebMock.disable_net_connect!
+    stub_request(:get, 'https://example.org:443/test')
+      .to_return(status: 200, body: 'ok', headers: { 'X-Zerocracy-Host' => 'evil.com' })
+    baza = BazaRb.new('example.org', 443, '000', loog: Loog::NULL)
+    home = baza.__send__(:home)
+    assert_equal(
```
Favixx, line 704: The assertion checks the less important of the two effects. `rehost` returns a modified `Iri`, but its damaging effect is the assignment `@host = host`; the return value is consumed only by the two chunk loops in `download` and `upload`, while every other method in the class rebuilds its URI from `home`, which reads `@host`. A refactor that assigned `@host` and then returned `uri` untouched would pass this assertion while leaving the client fully redirected for every non-chunked call.

Assert the side effect too: `assert_equal('example.org', baza.instance_variable_get(:@host))` after the call. The test already reaches into privates via `__send__(:rehost, ...)` and `__send__(:home)`, so one `instance_variable_get` costs nothing in consistency and pins the property that actually matters.

test/test_baza_edge.rb
```diff
@@ -695,6 +695,21 @@ def test_rehost_is_thread_safe
...
+      home.to_s,
+      baza.__send__(
+        :rehost,
+        Typhoeus::Request.get('https://example.org:443/test', headers: {}, connecttimeout: 1, timeout: 1), home
+      ).to_s
+    )
+  end
```
Favixx, line 711: The test that would justify the PR's security claim is missing: nothing here asserts that a hostile host is refused when the feature is enabled. The counterpart is what #355 is really asking for:

```ruby
def test_rehost_rejects_foreign_host
  baza = BazaRb.new('api.zerocracy.com', 443, '000', loog: Loog::NULL, sticky: true)
  home = baza.__send__(:home)
  ret = Typhoeus::Response.new(code: 200, headers: { 'X-Zerocracy-Host' => 'evil.com' })
  assert_equal(home.to_s, baza.__send__(:rehost, ret, home).to_s)
  assert_equal('api.zerocracy.com', baza.instance_variable_get(:@host))
end
```

Written today, that test fails: the client switches to `evil.com` without complaint. That is exactly the point — it is the executable statement of the requirement, and until it passes, #355 is not closed. Add it alongside the allowlist suggested on `rehost`, plus sibling cases for `localhost` and `169.254.169.254` so the internal-network and metadata-endpoint variants are pinned as well.

test/test_baza_edge.rb
```diff
@@ -840,7 +855,7 @@ def test_upload_switches_host_mid_chunks
       File.write(file, 'x' * 2_000_000)
       host = 'example.org'
       other = 'server2.example.org'
-      baza = BazaRb.new(host, 443, '000', loog: Loog::NULL, compress: false)
+      baza = BazaRb.new(host, 443, '000', loog: Loog::NULL, compress: false, sticky: true)
```
Favixx, line 858: These five `sticky: true` edits are not just test plumbing; after this PR they become the specification of the enabled mode, and that specification still says "trust any syntactically valid hostname". This case is the sharpest illustration: `test_upload_switches_host_mid_chunks` writes a 2 MB file, receives `X-Zerocracy-Host: server2.example.org` after the first chunk, and asserts that the remaining chunks go to the new host. Substitute `evil.com` for `server2.example.org` and the test still passes — meaning the suite currently certifies that an arbitrary host may receive the second half of a user's factbase along with `X-Zerocracy-Token`.

The download-side pair (`test_download_sticks_host`, `test_download_switches_host_mid_range`) certifies the same for the token plus the `Range` header on the retargeted GET. Once an allowlist exists, these fixtures should use hostnames inside the allowed suffix of the constructor host, so that the test data itself encodes the constraint. Right now they pass because there is no constraint at all, which means they will keep passing after an allowlist is added only by accident, and will not detect it being removed again.

PR #385, opened 2026-06-05
#357: Fix TOCTOU race in durable_place by opening file before checking

VasilevNStas:
## Problem

`durable_place` had a TOCTOU race: `File.exist?` check (line 306) happened before `File.open` (line 315). An attacker could swap the file path with a symlink between the check and the open.

## Solution

- Open the file first with `File.open(file, rb)` in block form (auto-closes)
- Check `f.stat.file?` on the opened file descriptor instead of `File.exist?`
- Use `f.size` instead of `File.size(file)` (same data, no re-check)
- Maintains backward-compatible error messages
- The `rescue Errno::ENOENT` preserves the absent-file error

## Verification
- [x] All tests pass

Closes #357.


GHX5T-SOL (changes requested):
I think this needs one more commit before merge.

The source change is in the right direction: `durable_place` now opens the path once and passes the same file descriptor into the upload instead of doing a separate `File.exist?`/`File.size`/`File.open` sequence. My local non-live validation passed:

- `bundle check`
- `ruby -c lib/baza-rb.rb`
- focused `test_durable_place` with coverage disabled (`1` test / `2` assertions)
- full `test/test_baza_edge.rb` with coverage disabled (`48` tests / `137` assertions)
- full `test/test_fake.rb` with coverage disabled (`40` tests / `53` assertions)
- `bundle exec rubocop` (`12` files, no offenses)
- `bundle exec rake picks`
- `bundle exec rake yard`

The blocker is coverage for the actual race fix. This PR changes only `lib/baza-rb.rb`; no test file changed. The existing `test_durable_place` is just a happy-path WebMock upload and would not fail on the old check/size/open sequence, so it does not protect the regression described in #357. Since this is a race/security fix, please add a focused local test that fails on the previous implementation and passes here, for example by swapping or replacing the path after the first existence/size check and asserting the posted `zip` still uses the originally opened file descriptor.

Also, the current hosted `rake` check is red on this head. The Ubuntu log I read fails in `TestBazaLive#test_live_full_cycle` with `Job #76594 ... did not finish`, which looks like live Zerocracy flakiness rather than the touched unit path, and I intentionally did not run live API tests locally. Still, the repo README asks contributors to submit with a green build, so please either rerun/fix the hosted rake job or document why this live failure is inherited/infrastructure-only before merge.

GHX5T-SOL (changes requested):
Rechecked current head `4da3e2be64c5f508f1dc3c722410390819138fdb`.

I think this still needs changes before merge.

First, the new regression is narrower than the risk described in `#357`. `test_durable_place_rejects_symlink_to_non_regular_file` uses a symlink to `/dev/null`, so the current code rejects it only because the opened target is a character device and `f.stat.file?` is false. A symlink to a regular file is still followed and uploaded.

I verified that locally without network traffic by overriding `post` and passing `durable_place` a symlink to a regular temp file; current head printed:

```text
id=42 uploaded="regular-target-content" symlink=true
```

That means the critical-file/symlink case described in `#357` is not covered by the new test and is still accepted if the path is a symlink to a regular file by the time `File.open(file, 'rb')` runs. If the intended fix is to reject symlink substitution, the implementation needs a no-follow/lstat-style guard and a regression that fails for a symlink to a regular file, not just `/dev/null`.

Second, local RuboCop is red on the new test:

```text
test/test_baza_edge.rb:52:3: C: Elegant/GoodMethodName: Method name "test_durable_place_rejects_symlink_to_non_regular_file" does not match the required pattern
test/test_baza_edge.rb:56:20: C: Style/FileNull: Use File::NULL instead of /dev/null.
test/test_baza_edge.rb:58:7: C: Elegant/NoRedundantVariable: Variable "error" is redundant and must be inlined
```

Validation I ran on the exact head:

- `ruby -c lib/baza-rb.rb` -> Syntax OK
- `ruby -c test/test_baza_edge.rb` -> Syntax OK
- focused new symlink test with coverage disabled -> `1` test / `3` assertions / `0` failures
- focused durable_place tests with coverage disabled -> `2` tests / `5` assertions / `0` failures
- full `test/test_baza_edge.rb` with coverage disabled -> `49` tests / `140` assertions / `0` failures
- full `test/test_fake.rb` with coverage disabled -> `40` tests / `53` assertions / `0` failures
- local no-network symlink-to-regular-file probe -> current head uploads the symlink target content
- `bundle exec rubocop` -> failed with the three offenses above
- `bundle exec rake yard` -> `96.23%` documented

Final live readback before this review showed the PR still open, non-draft, and on head `4da3e2be64c5f508f1dc3c722410390819138fdb`. I did not run full local `bundle exec rake` because it invokes live Zerocracy API tests; the hosted rake job is also red on the known live `TestBazaLive#test_live_full_cycle` failure while the non-rake hosted checks pass.

.rubocop.yml
```diff
@@ -30,6 +30,7 @@ Elegant/GoodMethodName:
     - format_value
     - retry_on
     - will_respond_with
+    - test_durable_place_rejects_symlink
```
yegor256: @VasilevNStas we don't need this, methods that start with `test_` are not flagged

lib/baza-rb.rb
```diff
@@ -315,26 +315,31 @@ def durable_place(pname, file)
...
+            'pname' => pname,
+            'file' => File.basename(file),
+            'zip' => f
+          }
+        ).headers['X-Zerocracy-DurableId'].to_i
+        throw(:"Durable ##{id} (#{file}, #{f.size} bytes) placed for job \"#{pname}\" at #{@host}")
+      end
     end
     id
+  rescue Errno::ENOENT
```
yegor256: @VasilevNStas what this is for? why rethrowing?

lib/baza-rb.rb
```diff
@@ -315,26 +315,31 @@ def durable_place(pname, file)
     raise(RuntimeError, "The name #{pname.inspect} is not valid") unless pname.match?(/\A[a-z0-9-]+\z/)
     raise(RuntimeError, "The name #{pname.inspect} is too long") if pname.length > 32
     raise(RuntimeError, 'The "file" of the durable is nil') if file.nil?
```
Favixx, line 318: Worth coordinating with #390, which is open against these exact lines. It replaces line 317 and the `File.exist?` this PR deleted with `valfile(file, must_exist: true)`, whose body is `raise ... if must_exist && !File.exist?(file)` followed by `raise ... if must_exist && File.empty?(file)` - two more independent path resolutions, landing immediately before the `File.open` on line 320.

If #390 merges after this one, the merged method resolves the same user-supplied path four times in a row: `lstat` (318), `exist?`, `empty?`, then `open` (320). The check-then-use pair this PR set out to remove would be reinstated one line above the open, by a helper that is meant to be adopted across every file-taking method in the class. The two PRs also conflict textually on 317-318, so one has to be rebased on the other regardless.

If the `NOFOLLOW`-on-open approach is taken, this call site needs `valfile(file, must_exist: false)` and the existence/emptiness verdicts have to come from the open descriptor (`Errno::ENOENT` and `f.size.zero?`). Otherwise the shared helper propagates this race into every method that adopts it, which is the opposite of what #390 is for.

lib/baza-rb.rb
```diff
@@ -315,26 +315,31 @@ def durable_place(pname, file)
...
-    raise(RuntimeError, "The file '#{file}' is absent") unless File.exist?(file)
-    if File.size(file) > 1024
-      raise(
-        RuntimeError,
-        "The file '#{file}' is too big (#{File.size(file)} bytes) for durable_place(), use durable_save() instead"
-      )
-    end
+    raise(RuntimeError, "The file '#{file}' is a symlink") if File.lstat(file).symlink?
```
Favixx: `File.lstat(file)` here, followed by `File.open(file, 'rb')` on line 320, is the same check-then-use shape the PR set out to remove: two independent resolutions of a user-supplied path with a schedulable gap between them. `rename(2)` is atomic, so an attacker who can write the containing directory swaps a symlink over `file` in that gap using exactly the primitive the original #357 report describes. The guard only rejects symlinks that were already in place when line 318 ran.

It is also incomplete as a symlink guard even without a race, because `lstat` inspects only the final path component. `durable_place('x', '/tmp/attacker/dir/f.bin')` where `dir` is a symlink passes this line, and `File.open` on 320 follows it anyway.

The atomic form pushes the check into the open itself: `File.open(file, File::RDONLY | File::BINARY | File::NOFOLLOW)`, rescuing `Errno::ELOOP` (Linux) and `Errno::EMLINK` (macOS/BSD - both OSes are in the matrix in `.github/workflows/rake.yml`) and raising the `is a symlink` RuntimeError from there. That removes this line entirely and leaves no window.

Separately, rejecting symlinks is a breaking change that the PR description does not mention: `durable_place` accepted symlinked paths in every released version. Symlinked artifact paths are routine in CI checkouts and build-output trees, and those callers get a `RuntimeError` from a patch upgrade.

lib/baza-rb.rb
```diff
@@ -315,26 +315,31 @@ def durable_place(pname, file)
...
     id = nil
```
Favixx: `BazaRb::Fake#durable_place` (`lib/baza-rb/fake.rb:145`) delegates to `checkfile`, which is `raise ... unless File.exist?(file)` then `raise ... unless File.size(file).positive?`. It has no symlink rejection, so after this PR the fake accepts an input the real client refuses. Any test or downstream consumer that exercises `durable_place` against `BazaRb::Fake` with a symlinked path passes while production raises - which is the failure mode a fake exists to prevent.

`checkfile` is also the same exist-then-size shape, so the fake keeps the pattern being removed from the real client here.

#390 is explicitly about making the two implementations reject the same set of inputs. Adding a real-only rejection widens the gap that PR is closing. Either mirror the guard in `checkfile`, or drop the guard in favour of `NOFOLLOW` at the open on line 320 - the fake never opens the file, so it would need no counterpart at all.

lib/baza-rb.rb
```diff
@@ -315,26 +315,31 @@ def durable_place(pname, file)
...
-      id = post(
-        home.append('durable-place'),
-        {
-          'pname' => pname,
-          'file' => File.basename(file),
-          'zip' => File.open(file, 'rb')
-        }
-      ).headers['X-Zerocracy-DurableId'].to_i
-      throw(:"Durable ##{id} (#{file}, #{File.size(file)} bytes) placed for job \"#{pname}\" at #{@host}")
+    File.open(file, 'rb') do |f|
```
Favixx: Moving the open ahead of the validation changes which exception class escapes for several inputs, and the change is platform-dependent. `File.open(dir, 'rb')` raises `Errno::EISDIR` on Linux, but succeeds on macOS (Darwin permits `O_RDONLY` on a directory), where control falls through to `f.stat.file?` on line 321. Both OSes are in the CI matrix.

Before this PR a directory took the `File.size(file) > 1024` branch on Linux (`st_size` is 4096) and produced a `RuntimeError`. Now it produces an `Errno::EISDIR`, which is not a `RuntimeError`, is not translated by the `rescue` on line 341, and is not in the `@raise` tag on this method. A unix socket behaves the same way via `Errno::ENXIO`.

Also note that `'rb'` as a mode string cannot carry `O_NOFOLLOW`. If you adopt the atomic symlink check suggested on line 318, this line has to become the integer-flags form `File.open(file, File::RDONLY | File::BINARY | File::NOFOLLOW)` anyway, so it is worth changing once.

lib/baza-rb.rb
```diff
@@ -315,26 +315,31 @@ def durable_place(pname, file)
...
+      raise(RuntimeError, "The file '#{file}' is absent") unless f.stat.file?
```
Favixx: `f.stat.file?` is not an absence check, and the message is false for everything it actually catches. If control reaches this line, `open(2)` on line 320 already succeeded, so the path demonstrably exists. The predicate is false only for non-regular files that opened successfully - a directory on macOS, `/dev/zero`, a FIFO with a writer attached - and for each of those, `The file '...' is absent` sends whoever reads the log hunting for a missing path that is right there.

The PR description calls this a replacement for `File.exist?`, but it is not equivalent in either direction: `File.exist?` returned true for directories and character devices, so those inputs used to reach the size branch or the upload, and now they get an absence error instead.

If the intent is `must be a regular file`, state that: `raise(RuntimeError, "The file '#{file}' is not a regular file") unless f.stat.file?`. Keep `is absent` for the `Errno::ENOENT` path on line 341, where it is the truth.

lib/baza-rb.rb
```diff
@@ -315,26 +315,31 @@ def durable_place(pname, file)
...
+      if f.size > 1024
```
Favixx: This guard is advisory only, and the PR does not change that. `f.size` is an `fstat(2)` on the descriptor from line 320, but the bytes that get uploaded come from a fresh `open(2)` that libcurl performs on the path (see the comment on line 334). The size that was validated and the size that is sent are measured on two different opens of the same name.

Concrete interleaving: `file` is a 10-byte regular file when this line runs, the guard passes, and before `curl_easy_perform` reaches the form the file is replaced by a 10 GB one. `durable_place` then streams 10 GB while having asserted it was under 1 KiB - the precise outcome this branch exists to prevent, and the reason the message on line 325 points at `durable_save`.

The pre-PR code had the same hole (`File.size(file)` then `File.open`), so this is not a regression - but the PR claims to close the check/use gap, and this is the branch where the gap has a measurable consequence. Since the contract is at most 1024 bytes, the check and the read can be a single operation: `bytes = f.read(1025)`, then `raise ... if bytes.bytesize > 1024`. One read, one truth, and `bytes` is what you upload.

lib/baza-rb.rb
```diff
@@ -315,26 +315,31 @@ def durable_place(pname, file)
...
+        raise(
+          RuntimeError,
+          "The file '#{file}' is too big (#{f.size} bytes) for durable_place(), use durable_save() instead"
+        )
+      end
+      elapsed(@loog, level: Logger::INFO) do
```
Favixx, line 342: Pulling `elapsed` inside the `File.open` block holds the descriptor open for the entire exchange, not just a read. `post` first calls `csrf` (line 546), a full synchronous `GET /csrf` round trip, then issues the POST; both sit under `attempt` (`with_retries(max_tries: @retries + 1, rescue: TimedOut)`, line 620) and `retry_on` (line 633), which loops on 429/5xx with `sleep(@pause * 2**attempt)`. With default retry and pause settings that is one descriptor pinned across potentially minutes of sleeping, per concurrent `durable_place`.

That cost would be justified if the descriptor were the thing being uploaded. It is not - libcurl re-opens the path on every attempt (line 334) - so the block buys the fd lifetime without buying the atomicity it was added for.

Worth flagging that this PR and #405 (`fix: #349 close 'zip' descriptor in BazaRb#durable_place`) are both rewriting these lines into a `File.open` block for different stated reasons and will conflict. Also relevant to that coordination: #349's premise that a retry re-sends an IO already at EOF is not what happens here, precisely because Ethon substitutes the path for the IO before libcurl ever sees it.

lib/baza-rb.rb
```diff
@@ -315,26 +315,31 @@ def durable_place(pname, file)
...
+        id = post(
+          home.append('durable-place'),
+          {
+            'pname' => pname,
+            'file' => File.basename(file),
+            'zip' => f
```
Favixx: This is the core problem: `'zip' => f` never causes the upload to read from the descriptor opened on line 320. `post` hands the hash to `Typhoeus::Request.post(..., body:)`; Typhoeus wraps it in `Ethon::Easy::Form`; `Ethon::Easy::Queryable#pairs_for` matches `when File, Tempfile` and replaces the IO with `file_info(v)`, which is `[File.basename(f.path), mime_type, File.expand_path(f.path)]`. `Ethon::Easy::Http::Postable#set_form` calls `form.materialize`, and `Form#form_add` passes `content[2]` - the path string - to `Curl.formadd` as `:form_option, :file`. libcurl opens that path itself when the transfer runs.

So check and use are still separated; they are separated by more code than before. Interleaving: `File.lstat` (318) and `File.open` (320) see a 10-byte regular file at `/tmp/x`; `f.stat.file?` (321) and `f.size` (322) pass; the attacker `rename(2)`s a symlink to `/etc/shadow` (or a huge file) over `/tmp/x`; `post` runs `csrf`, materialises the form, and libcurl `open(2)`s `/tmp/x`, follows the symlink and uploads the target. On a `TimedOut` or 429, `attempt`/`retry_on` re-yield and libcurl opens the path again, giving the attacker a fresh window per attempt.

Holding the IO cannot fix this, because Ethon exposes no in-memory multipart part (`CURLFORM_BUFFER` is not wired up in `Form#form_add`). Given the method already caps the payload at 1024 bytes, the atomic version is: open once with `File::NOFOLLOW`, `f.read(1025)`, enforce the size limit on the string, write it to a `Tempfile` under your own control, and hand libcurl that path. Then the bytes that were validated are provably the bytes that are sent.

lib/baza-rb.rb
```diff
@@ -315,26 +315,31 @@ def durable_place(pname, file)
...
+          }
+        ).headers['X-Zerocracy-DurableId'].to_i
+        throw(:"Durable ##{id} (#{file}, #{f.size} bytes) placed for job \"#{pname}\" at #{@host}")
```
Favixx: `f.size` here is a third `fstat(2)` on the same descriptor, taken after the upload rather than before it. If the file was truncated or rewritten while the POST was in flight, the logged size matches neither the value the guard on line 322 accepted nor what libcurl actually transferred, so the audit line contradicts the check that authorised it.

Capture it once - `size = f.size` above line 322 - and use that variable in the guard, in the message on line 325, and here. One syscall instead of three, and the number in the log is the number that was validated.

Line 325 has the same duplication: it calls `f.size` a second time to build a message it has already decided to raise.

lib/baza-rb.rb
Favixx: A `rescue` at method-body level covers everything from line 313 down, which here includes `csrf`, the POST, `retry_on`'s sleep loop, and Typhoeus/Ethon's FFI work. Any `Errno::ENOENT` raised anywhere beneath `post` - a missing CA bundle path, a `Dir` or `File` operation inside a dependency, a unix-socket path if one is ever configured - is relabelled `The file '<user's path>' is absent`, pointing the reader at a path that was never the problem. Debugging that from a production log would be genuinely hard, because the message names a file that exists.

Only two calls in this method can legitimately raise `ENOENT`: `File.lstat` on line 318 and `File.open` on line 320. Scope the rescue to them - wrap 318-320 in a `begin ... rescue Errno::ENOENT ... end`, or drop line 318 per the `NOFOLLOW` suggestion and let the open be the single source of the absence verdict.

The re-raise also drops the errno's own path from the message, which is exactly the detail that distinguishes `your file is gone` from `something else is gone`. `e.message` is worth including.

lib/baza-rb.rb
```diff
@@ -315,26 +315,31 @@ def durable_place(pname, file)
...
+      end
     end
     id
+  rescue Errno::ENOENT
+    raise(RuntimeError, "The file '#{file}' is absent")
```
Favixx: Only `ENOENT` is translated, so the other errnos `File.lstat` (line 318) can raise now escape as `SystemCallError`, where the pre-PR `File.exist?` returned `false` and produced a uniform `RuntimeError, "The file '...' is absent"`:

- `Errno::EACCES` - a parent directory without the execute bit. Routine under CI sandboxes and Docker `USER` switches.
- `Errno::ENOTDIR` - a non-directory component in the path.
- `Errno::ENAMETOOLONG` and `Errno::ELOOP` - long or looping paths.

Every other validation failure in this method raises `RuntimeError`, and the tests this PR deletes asserted exactly that class. Callers wrapping `durable_place` in `rescue RuntimeError` stop catching these cases silently, on a patch upgrade.

If the pre-PR contract is meant to hold, rescue `SystemCallError` and translate. If the contract is deliberately changing, the RDoc needs to say so - the `@raise` tag on this method still lists only `ServerFailure`, and now `Errno::EISDIR` (line 320), `Errno::EACCES` and `Errno::ENOTDIR` can all come out of it.

lib/baza-rb.rb
```diff
@@ -315,26 +315,31 @@ def durable_place(pname, file)
...
   end
 
   # Save a single durable from local file to server.
```
Favixx, line 357: The sibling methods still carry the pattern this PR is removing. `durable_save` on line 356 is `raise ... unless File.exist?(file)`, then `upload(...)`, which does `File.size(file)` on line 863 and `File.open(file, 'rb')` on line 871 - four independent resolutions of the same user-supplied path. Worse, that `File.open` is inside `upload`'s `loop`, so a chunked transfer re-resolves the path once per chunk and once per rewind, giving an attacker a window per chunk rather than one per call. `durable_load` (line 375) calls `File.size(file)` on a path `download` has just finished writing.

This matters for scope: the error message on line 325 tells users that anything over 1 KiB belongs in `durable_save`, so the method with the widest window and the longest-lived path reference is the one left unfixed. Fixing the small-file path while the large-file path keeps the same defect inverts the risk ordering.

If narrowing to `durable_place` is deliberate, that is defensible - but #357 is being closed by this PR, so the sibling cases should be filed rather than silently dropped.

test/test_baza_edge.rb
```diff
@@ -49,22 +49,20 @@ def test_durable_place
...
-  def test_durable_place_raises_when_pname_is_too_long
-    assert_includes(
-      assert_raises(RuntimeError) { fake_baza.durable_place('a' * 33, '/tmp/x') }.message,
-      'is too long'
-    )
-  end
-
-  def test_durable_find_raises_when_pname_is_invalid
-    assert_includes(assert_raises(RuntimeError) { fake_baza.durable_find('BAD!', 'file') }.message, 'is not valid')
+  def test_durable_place_rejects_symlink
```
Favixx, line 70: This test does not replace what it removes. Three passing tests are deleted here: `test_durable_place_raises_when_pname_is_invalid`, `test_durable_place_raises_when_pname_is_too_long`, and `test_durable_find_raises_when_pname_is_invalid`. None of them relate to #357 - the `pname` guards on lines 315-316 of `lib/baza-rb.rb` are untouched by this PR, and `durable_find` is not modified at all.

Nothing else in the suite covers them. After this change, `grep -rn 'durable_place\|durable_find' test/` leaves no assertion on either method's name validation, while the byte-for-byte equivalent assertions for `lock`, `unlock`, `name_exists?`, `recent` and `push` remain in this same file at lines 704-735. That is an inconsistency inside one file, so the next person reading it will reasonably conclude those two methods do not validate their names.

`test__helper.rb` sets `SimpleCov.minimum_coverage(90)` and `minimum_coverage_by_file(95)`, so the deletions also consume the headroom the new branches need - which makes it hard to tell from a green build whether the new code is actually covered.

Please restore the three tests and add this one alongside them rather than in their place.

test/test_baza_edge.rb
```diff
@@ -49,22 +49,20 @@ def test_durable_place
...
+    WebMock.disable_net_connect!
+    Dir.mktmpdir do |dir|
+      real = File.join(dir, 'real.bin')
+      File.binwrite(real, 'content')
+      link = File.join(dir, 'link.bin')
+      File.symlink(real, link)
```
Favixx: `File.symlink` raises `NotImplementedError` on Windows and `Errno::EPERM` on Windows without Developer Mode or `SeCreateSymbolicLinkPrivilege`, and it fails on filesystems with no symlink support (some container overlay and network mounts). The matrix in `.github/workflows/rake.yml` is `[ubuntu-24.04, macos-15]` so this is green today, but the gemspec does not restrict the platform and this becomes the only test in the suite whose outcome depends on a filesystem capability.

A guard keeps it honest without weakening it:

```ruby
begin
  File.symlink(real, link)
rescue NotImplementedError, Errno::EPERM
  skip('symlinks are not supported on this platform')
end
```

The test would also be stronger with the negative half asserted in the same `Dir.mktmpdir`: a plain regular file still succeeds. Right now nothing pins that the new `lstat` guard cannot regress into rejecting everything - `test_durable_place` happens to use a non-symlinked path, but for unrelated reasons and in a different fixture.

test/test_baza_edge.rb
```diff
@@ -49,22 +49,20 @@ def test_durable_place
...
+      assert_includes(
+        assert_raises(RuntimeError) do
+          fake_baza(compress: false).durable_place('simple', link)
+        end.message,
+        'symlink'
+      )
+    end
   end
```
Favixx, line 111: The suite now asserts that a symlink which already existed is rejected - the one property the fix establishes without needing a race at all. Nothing here exercises the behaviour #357 is about, and nothing exercises the descriptor lifetime that motivated moving to the block form.

Two tests would actually pin this PR's claims:

1. Descriptor closure. Stub the POST to raise, call `durable_place`, and assert the handle is gone: `ObjectSpace.each_object(File).count { |h| !h.closed? && h.path == file }` around the call, or on Linux compare `Dir.children('/proc/self/fd').size` before and after. This is the single property `File.open`-with-block buys over the old code, it is currently untested, and it matters because #405 is landing the same structural change for #349 - without a test, whichever merges second looks like a no-op.

2. Retry behaviour. Stub the POST to time out once and then succeed (the pattern already used elsewhere in this file) and assert the second attempt carries the file bytes rather than an empty part. That pins what `attempt` / `retry_on` do with the params hash across attempts, which is the disputed point in #349.

Neither can prove a race is closed - that is fair, races are not unit-testable - but both pin the mechanism this PR introduces, which `assert_includes(..., 'symlink')` does not.

PR #389, opened 2026-06-05
#362: add BazaRb::Error hierarchy for custom errors

VasilevNStas:
## Problem

All custom errors (`ServerFailure`, `TimedOut`, `BadCompression`) inherited from `StandardError` directly. All validation errors used `raise(RuntimeError, ...)`. Callers had no common ancestor to catch all BazaRb errors.

## Solution

- Added `BazaRb::Error < StandardError` as the base class for all library errors
- `ServerFailure`, `TimedOut`, `BadCompression` now inherit from `Error` instead of `StandardError`
- `ConnectionFailed` still inherits from `TimedOut` (unchanged)
- All validation errors (`raise(RuntimeError, ...)`) changed to `raise(Error, ...)` in both `baza-rb.rb` and `fake.rb`
- All test assertions updated from `assert_raises(RuntimeError)` to `assert_raises(BazaRb::Error)`

Now callers can `rescue BazaRb::Error` to catch any error from this library, or be more specific with individual subclasses.

```ruby
begin
  baza.push(...)
rescue BazaRb::ValidationError # future
  # handle bad input
rescue BazaRb::ServerFailure
  # handle server error
rescue BazaRb::Error
  # catch anything else from BazaRb
end
```

## Verification
- [x] All tests pass (45 edge, 40 fake, 26 pact)
- [x] Verified the complete hierarchy: `BazaRb::Error < StandardError`, `ServerFailure < Error`, `TimedOut < Error`, `ConnectionFailed < TimedOut`, `BadCompression < Error`

Closes #362.


lib/baza-rb.rb
```diff
@@ -834,10 +839,10 @@ def download(uri, file)
         @loog.debug(msg.compact.join(', '))
         break if ret.code == 200
         range, total = crange(rheaders)
-        raise(RuntimeError, "Total size is not valid (#{total.inspect})") unless total.match?(/\A(?:\*|[0-9]+)\z/)
+        raise(Error, "Total size is not valid (#{total.inspect})") unless total.match?(/\A(?:\*|[0-9]+)\z/)
         _b, e = range.split('-', 2)
-        raise(RuntimeError, "Range is not valid (#{range.inspect})") if e.nil?
-        raise(RuntimeError, "Range is not valid (#{range.inspect})") unless e.match?(/\A[0-9]+\z/)
+        raise(Error, "Range is not valid (#{range.inspect})") if e.nil?
+        raise(Error, "Range is not valid (#{range.inspect})") unless e.match?(/^[0-9]+$/)
```
morphqdd: This changes the anchoring from `/\A[0-9]+\z/` to `/^[0-9]+$/`, which is unrelated to the error-hierarchy goal of this PR and is a regression: in Ruby `^`/`$` match line boundaries, not string boundaries, so a value like `"12\nfoo"` now passes this guard. Note that line 842 in the same method still uses `\A...\z`, so the method is now internally inconsistent. The existing `test_download_rejects_range_end_with_newline` does not catch this because it fails earlier inside `crange` (whitespace `split` drops the newline), so this weakening is untested. Please revert to `/\A[0-9]+\z/`.

lib/baza-rb.rb
```diff
@@ -509,14 +512,16 @@ def fee(tab, amount, summary, job)
...
   def enter(pname, badge, why, job)
-    raise(RuntimeError, 'The "pname" is nil') if pname.nil?
-    raise(RuntimeError, 'The "pname" may not be empty') if pname.empty?
-    raise(RuntimeError, 'The "badge" is nil') if badge.nil?
-    raise(RuntimeError, 'The "badge" may not be empty') if badge.empty?
-    raise(RuntimeError, 'The "why" is nil') if why.nil?
-    raise(RuntimeError, 'The "why" may not be empty') if why.empty?
-    raise(RuntimeError, 'The "job" must be an Integer') unless job.nil? || job.is_a?(Integer)
-    raise(RuntimeError, 'The "job" must be positive') unless job.nil? || job.positive?
+    raise(Error, 'The "pname" is nil') if pname.nil?
```
morphqdd: These two `pname` validations (`is not valid` / `is too long`) are new behavior that did not exist on `master` and is unrelated to #362 (error hierarchy). Bundling a behavior change into an error-class rename makes the PR harder to review and to revert. Could you move these into a separate PR?

lib/baza-rb.rb
```diff
@@ -509,14 +512,16 @@ def fee(tab, amount, summary, job)
...
+    raise(Error, 'The "pname" may not be empty') if pname.empty?
```
morphqdd: If these new `enter` validations stay, they need their own tests — currently nothing exercises the `is not valid` / `is too long` paths for `enter`.

lib/baza-rb.rb
```diff
@@ -107,14 +110,14 @@ def balance
...
   def push(pname, data, meta, chunk_size: DEFAULT_CHUNK_SIZE)
-    raise(RuntimeError, 'The "name" of the job is nil') if pname.nil?
-    raise(RuntimeError, 'The "name" of the job may not be empty') if pname.empty?
-    raise(RuntimeError, "The name #{pname.inspect} is not valid") unless pname.match?(/\A[a-z0-9-]+\z/)
-    raise(RuntimeError, "The name #{pname.inspect} is too long") if pname.length > 32
-    raise(RuntimeError, 'The "data" of the job is nil') if data.nil?
-    raise(RuntimeError, 'The "data" of the job may not be empty') if data.empty?
-    raise(RuntimeError, 'The "meta" of the job is nil') if meta.nil?
-    raise(RuntimeError, 'The "meta" of the job must be an Array') unless meta.is_a?(Array)
+    raise(Error, 'The "name" of the job is nil') if pname.nil?
```
morphqdd: All argument validations collapse into the generic `Error`. The PR description itself promises a `BazaRb::ValidationError` in its `rescue` example, but it does not exist. Introducing `ValidationError < Error` and using it for these input-validation raises would make that example real and give the hierarchy actual meaning (bad-input vs server/transport failures), which is the useful distinction a caller wants to `rescue` on.

lib/baza-rb.rb
```diff
@@ -227,18 +230,18 @@ def verified(id)
   # @raise [RuntimeError] If the name is already locked
```
morphqdd: This `@raise [RuntimeError]` tag (and the many `@raise [ServerFailure]` tags above validation blocks throughout the file) are now inaccurate — the code raises `BazaRb::Error`. Please update the YARD tags to match, otherwise the docs drift from behavior.

lib/baza-rb.rb
```diff
@@ -35,19 +35,22 @@
 class BazaRb
   DEFAULT_CHUNK_SIZE = 1_000_000
 
+  # Base class for all BazaRb errors.
+  class Error < StandardError; end
```
morphqdd: Nice base class. It would be worth a small test that locks the contract this PR is selling — e.g. `assert_operator(BazaRb::ServerFailure, :<, BazaRb::Error)` for each subclass — so a future refactor cannot silently break `rescue BazaRb::Error`.

lib/baza-rb.rb
```diff
@@ -35,19 +35,22 @@
...
+
   # When the server failed (503).
-  class ServerFailure < StandardError; end
+  class ServerFailure < Error; end
```
morphqdd: Worth calling out in the CHANGELOG (and a version bump): `BazaRb::Error` is not a `RuntimeError`, so any downstream code doing `rescue RuntimeError` to catch these validation errors will silently stop catching them after this change.

lib/baza-rb/fake.rb
```diff
@@ -38,9 +38,9 @@ def whoami
   # @return [Integer] Always returns 42 as the fake job ID
   def push(name, data, meta, chunk_size: BazaRb::DEFAULT_CHUNK_SIZE) # rubocop:disable Lint/UnusedMethodArgument
     checkname(name)
```
morphqdd, line 40: Style inconsistency: `fake.rb` uses the fully-qualified `BazaRb::Error` while `baza-rb.rb` uses the bare `Error`. Since `Fake` is nested inside `BazaRb`, the bare `Error` resolves here too — please pick one form for consistency across the codebase.

test/test_baza_edge.rb
```diff
@@ -51,20 +51,20 @@ def test_durable_place
 
   def test_durable_place_raises_when_pname_is_invalid
     assert_includes(
-      assert_raises(RuntimeError) { fake_baza.durable_place('INVALID', '/tmp/x') }.message,
+      assert_raises(BazaRb::Error) { fake_baza.durable_place('INVALID', '/tmp/x') }.message,
```
morphqdd: These assertions only swap the expected class from `RuntimeError` to `BazaRb::Error`, but nothing verifies the new hierarchy itself (that a subclass is caught by rescuing the base). Consider one test asserting the ancestry, as noted on the base-class definition.

lib/baza-rb.rb
```diff
@@ -38,19 +38,25 @@ class BazaRb
 
   include Compress
 
+  # Base class for all BazaRb errors.
+  class Error < StandardError; end
```
Favixx: Worth weighing a module here instead of a base class, because it gets you the same `rescue` ergonomics without the semver break.

Ruby's `rescue` dispatches through `Module#===`, not through `Class#<`, so a bare module works as a rescue target for any exception class that includes it. That means:

```ruby
module Error; end
class ValidationError < RuntimeError; include Error; end
class ServerFailure  < StandardError; include Error; end
class TimedOut       < StandardError; include Error; end
class ConnectionFailed < TimedOut; end
class BadCompression < StandardError; include Error; end
```

With this shape `rescue BazaRb::Error` catches everything the library throws — the goal of #362 — while `ValidationError` stays a `RuntimeError`, so any existing caller doing `rescue RuntimeError` around `push`/`pull`/`transfer` keeps working. Under the current class-based shape that caller silently stops catching, since `Error < StandardError`.

The cost is that `BazaRb::Error.new('x')` is no longer instantiable, which is arguably a feature given the six places below that currently `raise(Error, ...)` directly. The counter-argument for keeping a class is precedent: `zerocracy/fbe` already does `Fbe::Error < StandardError` in `lib/fbe.rb`, and consistency across the org has value. I'd accept either, but the trade-off should be a deliberate choice recorded in the PR body rather than left implicit.

lib/baza-rb.rb
```diff
@@ -38,19 +38,25 @@ class BazaRb
...
+
+  # When an invalid argument is provided.
+  class ValidationError < Error; end
```
Favixx: This line is the actual breaking change in the PR, and nothing in the description mentions it. `ValidationError < Error < StandardError` means it is no longer a `RuntimeError`, so `begin; baza.push(...); rescue RuntimeError => e; ...; end` stops catching every validation failure in the library. `rescue => e` still catches, which makes the regression silent rather than loud.

I checked how much this matters in practice before flagging it. `gh api 'search/code?q=org:zerocracy+"rescue+RuntimeError"'` returns `total_count: 0`. `judges-action` has `RuntimeError` in `lib/qos_search.rb` (a `raise`) and `test/lib/test_supervision.rb` (an unrelated `assert_raises`), neither near a baza call. `swarm-template` has baza-rb only in `Gemfile.lock`. `fbe/lib/fbe/enter.rb` calls `baza.enter` but raises/rescues its own `Fbe::Error`. So no in-org consumer breaks.

baza.rb is published on rubygems, though, so external callers exist that I cannot audit. Please note the incompatibility explicitly in the PR body so it lands in the release notes, or adopt the module form suggested on line 42, which avoids the break entirely.

Separately, the PR body still says "All validation errors ... changed to `raise(Error, ...)`" and shows `rescue BazaRb::ValidationError # future`. The code has moved past that; please refresh the description so the merge commit message is accurate.

lib/baza-rb.rb
```diff
@@ -38,19 +38,25 @@ class BazaRb
...
+
   # When the server failed (503).
-  class ServerFailure < StandardError; end
+  class ServerFailure < Error; end
```
Favixx: Merge-order note, since these exact lines are rewritten by your own PR #390 in this repo.

#390 ("#368/#369: shared Validation module + contract test") touches `lib/baza-rb.rb`, `lib/baza-rb/fake.rb`, `test/test_baza_edge.rb` and `test/test_fake.rb` and replaces every `raise(RuntimeError, ...)` with `raise(BazaRb::ValidationError, ...)` or a `valname`/`valid`/`valfile`/`valowner` call — but it declares `class ValidationError < StandardError` directly, with no `Error` base and with `ServerFailure < StandardError` left as-is. That is not just a textual conflict across ~100 lines; the two PRs assert different hierarchies for the same constant.

Concretely: if #390 merges after #389, the second diff hunk reintroduces `ValidationError < StandardError` and drops `Error` entirely, and `test_error_hierarchy` added below starts failing (or, worse, `Error` survives unreferenced and `rescue BazaRb::Error` silently stops catching validation failures).

This PR should land first, and #390 should then be rebased to delete its own `class ValidationError < StandardError` line and its `class ServerFailure < StandardError` line, keeping the hierarchy defined here. Please add that dependency note to #390 so whoever merges does it in the right order.

lib/baza-rb.rb
```diff
@@ -110,14 +116,14 @@ def balance
   # @param [Integer] chunk_size Maximum size of one chunk
   # @raise [ServerFailure] If the push operation fails
```
Favixx, line 127: The `@raise` tags were not updated. `push` now raises `ValidationError` in eight places (lines 119-126), but the only documented failure mode is still `@raise [ServerFailure] If the push operation fails`. A user reading the generated YARD has no way to discover `BazaRb::ValidationError` exists, which defeats the point of naming it.

This is systematic rather than a one-off: the same stale-or-missing situation holds for `pull` (147), `finished?` (167), `stdout` (185), `exit_code` (202), `verified` (219), `unlock` (258), `recent` (276), `durable_place` (317), `durable_save` (351), `durable_load` (368), `durable_lock` (384), `durable_unlock` (401), `transfer` (448), `fee` (484), `enter` (526). `name_exists?` and `durable_find` have no `@raise` tag at all despite four and six validation raises respectively. Only `lock` (236) was touched.

Suggest adding `# @raise [ValidationError] If any argument is invalid` above each of these, next to the existing `@raise [ServerFailure]` line. It is mechanical, and it makes the new class discoverable from the docs rather than only from the source.

lib/baza-rb.rb
```diff
@@ -227,21 +233,21 @@ def verified(id)
   #
   # @param [String] pname The name of the product on the server
   # @param [String] owner The owner of the lock (any string)
-  # @raise [RuntimeError] If the name is already locked
+  # @raise [Error] If the name is already locked
```
Favixx: This tag went from `@raise [RuntimeError]` to `@raise [Error]`, which is technically accurate for line 250 but strictly less informative than before. "The name is already locked" is a specific, expected, recoverable outcome — arguably the single most important error this method can produce, since the caller is going to want to back off and retry or skip.

As written, the only way to detect it is `rescue BazaRb::Error` plus a message match on `"it's already locked"`, which is exactly the message-parsing that issue #362 set out to remove: "Callers cannot distinguish error types ... Must parse `e.message`, which may change."

A dedicated class fixes both the tag and the caller experience:

```ruby
# When a product is already locked by someone else.
class Locked < Error; end
```

then `@raise [Locked] If the name is already locked by another owner`. That also stops this case from being conflated with the `download`/`crange` protocol errors, which currently share the same `Error` class.

lib/baza-rb.rb
```diff
@@ -227,21 +233,21 @@ def verified(id)
...
+    raise(ValidationError, "The name #{pname.inspect} is too long") if pname.length > 32
+    raise(ValidationError, 'The "owner" of the lock is nil') if owner.nil?
+    raise(ValidationError, 'The "owner" of the lock may not be empty') if owner.empty?
     elapsed(@loog, level: Logger::INFO) do
       throw(:"Product name #{pname.inspect} locked at #{@host}") if post(
         home.append('lock').append(pname),
         { 'owner' => owner }, [302, 409]
       ).code == 302
-      raise(RuntimeError, "Failed to lock #{pname.inspect} product at #{@host}, it's already locked")
+      raise(Error, "Failed to lock #{pname.inspect} product at #{@host}, it's already locked")
```
Favixx: Raising the abstract base directly. `Error` is documented on line 41 as "Base class for all BazaRb errors"; if it is also a concrete throwable, then `rescue BazaRb::Error` cannot tell "the product is locked by someone else" (a normal, expected outcome) from "you passed a nil pname" (a bug in the caller) from "the server sent a malformed Content-Range" (a server bug).

Concrete failure: a caller that wants lock contention to be non-fatal has to write

```ruby
rescue BazaRb::Error => e
  raise unless e.message.include?("it's already locked")
```

which breaks the moment anyone rewords the message — and the message is interpolated with `@host`, so it is not stable to match on either.

There are nine such `raise(Error, ...)` sites in this file (250, 324, 357, 835, 837, 838, 850, 852, 854), in six groups covering three genuinely different categories. Suggest `Locked < Error` here, `ValidationError` at 324/357 (see those lines), and a `ProtocolError < Error` for the `download`/`crange` group, leaving `Error` as a pure rescue target that is never instantiated.

lib/baza-rb.rb
```diff
@@ -310,15 +316,15 @@ def name_exists?(pname)
...
-    raise(RuntimeError, "The name #{pname.inspect} is not valid") unless pname.match?(/\A[a-z0-9-]+\z/)
-    raise(RuntimeError, "The name #{pname.inspect} is too long") if pname.length > 32
-    raise(RuntimeError, 'The "file" of the durable is nil') if file.nil?
-    raise(RuntimeError, "The file '#{file}' is absent") unless File.exist?(file)
+    raise(ValidationError, 'The "pname" of the durable is nil') if pname.nil?
+    raise(ValidationError, 'The "pname" of the durable may not be empty') if pname.empty?
+    raise(ValidationError, "The name #{pname.inspect} is not valid") unless pname.match?(/\A[a-z0-9-]+\z/)
+    raise(ValidationError, "The name #{pname.inspect} is too long") if pname.length > 32
+    raise(ValidationError, 'The "file" of the durable is nil') if file.nil?
+    raise(Error, "The file '#{file}' is absent") unless File.exist?(file)
```
Favixx: This introduces a Fake/Real divergence that did not exist before this PR, and it is the same class of bug that #368/#369 and your PR #390 exist to eliminate.

On `master` both sides raised `RuntimeError` for a missing durable file, so a test written against `BazaRb::Fake` was a faithful proxy for production. After this PR: `BazaRb::Fake#durable_place` calls `checkfile`, which at `lib/baza-rb/fake.rb:283` raises `BazaRb::ValidationError, 'The file must exist'`; the real `BazaRb#durable_place` raises the bare `Error` here. So this passes in test and blows up in production:

```ruby
begin
  baza.durable_place('proj', '/no/such/file')
rescue BazaRb::ValidationError
  # green against Fake, uncaught against the real client
end
```

A nonexistent path is an invalid argument, so `ValidationError` is the right class on both sides. That is also what your own #390 lands on — `valfile(file, must_exist: true)` raises `BazaRb::ValidationError, "The file '#{file}' is absent"` — so this line will be reverted to `ValidationError` one PR later anyway. Please just use `ValidationError` here (and at 357) so the two PRs agree and the Fake stays truthful.

lib/baza-rb.rb
```diff
@@ -344,11 +350,11 @@ def durable_place(pname, file)
...
-    raise(RuntimeError, 'The ID of the durable is nil') if id.nil?
-    raise(RuntimeError, 'The ID of the durable must be an Integer') unless id.is_a?(Integer)
-    raise(RuntimeError, 'The ID of the durable must be a positive integer') unless id.positive?
-    raise(RuntimeError, 'The "file" of the durable is nil') if file.nil?
-    raise(RuntimeError, "The file '#{file}' is absent") unless File.exist?(file)
+    raise(ValidationError, 'The ID of the durable is nil') if id.nil?
+    raise(ValidationError, 'The ID of the durable must be an Integer') unless id.is_a?(Integer)
+    raise(ValidationError, 'The ID of the durable must be a positive integer') unless id.positive?
+    raise(ValidationError, 'The "file" of the durable is nil') if file.nil?
+    raise(Error, "The file '#{file}' is absent") unless File.exist?(file)
```
Favixx: Same divergence as line 324. `BazaRb::Fake#durable_save` calls `checkfile`, which raises `BazaRb::ValidationError` for a missing file (`lib/baza-rb/fake.rb:283`), while the real implementation raises bare `Error` here.

Note also the internal inconsistency inside this single method: line 356 raises `ValidationError` for `file.nil?`, and line 357 raises `Error` for a file that is merely absent. Both are the same kind of failure — the caller handed over an unusable path — so a caller has to catch two different classes to cover one argument.

Making both `ValidationError` also matches `valfile` in `lib/baza-rb/validation.rb` from #390, which raises `BazaRb::ValidationError` for nil, empty, absent and empty-content in one place. Aligning now removes a conflict from that rebase instead of adding one.

lib/baza-rb.rb
```diff
@@ -441,18 +447,20 @@ def durable_find(pname, file)
   # @return [Integer] Receipt ID for the transaction
   # @raise [ServerFailure] If the transfer fails
   def transfer(recipient, amount, summary, job: nil, badge: nil)
-    raise(RuntimeError, 'The "recipient" is nil') if recipient.nil?
-    raise(RuntimeError, "The recipient #{recipient.inspect} is not valid") unless recipient.match?(/\A[a-zA-Z0-9-]+\z/)
-    raise(RuntimeError, 'The "amount" is nil') if amount.nil?
+    raise(ValidationError, 'The "recipient" is nil') if recipient.nil?
+    unless recipient.match?(/\A[a-zA-Z0-9-]+\z/)
```
Favixx, line 478: This reflow from a one-line guard to a three-line `unless` block is not required by the error-class change — it is required only because `ValidationError` is 11 characters longer than `RuntimeError` and the original line was already near the 120-column limit. Same pattern at `lib/baza-rb/fake.rb:205`.

The reason it is worth mentioning is collision cost. PR #390 replaces this exact block: it deletes the inline recipient regex and moves the check elsewhere. So this hunk is guaranteed to conflict, and the conflict resolution will be "take #390's version", meaning the reformat is churn that only exists to be thrown away.

A cheaper way to keep the guard on one line and dodge the conflict:

```ruby
VE = ValidationError
...
raise(VE, "The recipient #{recipient.inspect} is not valid") unless recipient.match?(/\A[a-zA-Z0-9-]+\z/)
```

or leave the reflow but flag it in the PR body as expected-to-conflict so the person rebasing #390 does not try to preserve it.

lib/baza-rb.rb
```diff
@@ -824,10 +832,10 @@ def download(uri, file)
         @loog.debug(msg.compact.join(', '))
         break if ret.code == 200
         range, total = crange(rheaders)
-        raise(RuntimeError, "Total size is not valid (#{total.inspect})") unless total.match?(/\A(?:\*|[0-9]+)\z/)
+        raise(Error, "Total size is not valid (#{total.inspect})") unless total.match?(/\A(?:\*|[0-9]+)\z/)
```
Favixx: This is a server-side protocol violation — the server returned a 206 with a `Content-Range` whose total is not `*` or a number — and it is being reported with the same class as "you passed a nil job name".

That is a real discrimination loss for callers. Someone writing retry/alerting logic wants `rescue BazaRb::ServerFailure` (or an equivalent "the remote misbehaved" class) to cover it, because the correct response is to retry against another host or page the server team, not to fix the call site. Under this PR `rescue BazaRb::ServerFailure` misses it entirely, and `rescue BazaRb::Error` also swallows their own argument bugs.

Worth noting these three raises (835, 837, 838) sit inside the `elapsed` block but *outside* the `attempt` wrapper, so unlike `TimedOut` they are not retried — which is another reason they deserve a name distinct from validation failures.

Suggest either `ServerFailure` here, or a new `ProtocolError < Error` shared with the three raises in `crange` below. The `test_download_rejects_malformed_total_size` and `test_download_rejects_range_end_with_newline` assertions in `test/test_baza_edge.rb` would then name that class instead of the base.

lib/baza-rb.rb
```diff
@@ -839,11 +847,11 @@ def download(uri, file)
 
   def crange(headers)
     crange = headers['Content-Range']
-    raise(RuntimeError, 'Content-Range header is missing') if crange.nil?
+    raise(Error, 'Content-Range header is missing') if crange.nil?
```
Favixx: Same category as line 835: a missing or malformed `Content-Range` header is the server violating RFC 7233, not the caller passing bad input, yet it lands in the same `Error` bucket as `push(nil, ...)`.

All three raises in this method (850, 852, 854) plus the three in `download` describe one condition — "the range response is unusable" — and would be better served by one named class, e.g. `ProtocolError < Error`, with `@raise [ProtocolError]` on `download`.

A further nit while you are in here: 852 and 854 produce byte-identical messages (`"Content-Range is not valid (#{crange.inspect})"`) for two different structural failures — no space-separated value, versus no `/`-separated total. Since the whole point of this PR is to stop callers from parsing messages, and these two cannot be told apart by class *or* by message, distinguishing them (`"...has no value"` / `"...has no total"`) would be a small win for anyone debugging a real 206 from the server.

lib/baza-rb/fake.rb
```diff
@@ -38,9 +38,9 @@ def whoami
...
-    raise(RuntimeError, 'The "data" of the job is nil') if data.nil?
-    raise(RuntimeError, 'The data must be non-empty') if data.empty?
-    raise(RuntimeError, 'The meta must be an array') unless meta.is_a?(Array)
+    raise(BazaRb::ValidationError, 'The "data" of the job is nil') if data.nil?
```
Favixx, line 41: Correcting a point from the earlier review round: the fully-qualified `BazaRb::ValidationError` here is not a style choice, it is required, and switching to the bare `ValidationError` to match `lib/baza-rb.rb` would raise `NameError` at runtime.

The reason is the compact definition on line 24 of this file, `class BazaRb::Fake`. With the compact form, `Module.nesting` is `[BazaRb::Fake]` — `BazaRb` is *not* in the lexical scope — so constant lookup for a bare `ValidationError` searches `BazaRb::Fake` and then its ancestors (`Object`), never `BazaRb`. It would only work if the file were rewritten as nested `class BazaRb` / `class Fake`.

So the asymmetry between the two files is unavoidable as long as `fake.rb` uses the compact form. If the inconsistency bothers you, the fix is to qualify in `baza-rb.rb` too (`raise(BazaRb::ValidationError, ...)`), which is what your PR #390 does throughout — that would also make the two PRs textually closer and cut down the rebase conflict.

lib/baza-rb/fake.rb
```diff
@@ -201,17 +201,19 @@ def balance
...
     end
-    raise(RuntimeError, 'The "amount" must be positive') unless amount.positive?
-    raise(RuntimeError, 'The "summary" is nil') if summary.nil?
-    raise(RuntimeError, "The summary #{summary.inspect} is empty") if summary.empty?
+    raise(BazaRb::ValidationError, 'The "amount" must be positive') unless amount.positive?
+    raise(BazaRb::ValidationError, 'The "summary" is nil') if summary.nil?
+    raise(BazaRb::ValidationError, "The summary #{summary.inspect} is empty") if summary.empty?
     checkid(job) unless job.nil?
-    raise(RuntimeError, 'The "badge" must be a String') if !badge.nil? && !badge.is_a?(String)
+    raise(BazaRb::ValidationError, 'The "badge" must be a String') if !badge.nil? && !badge.is_a?(String)
```
Favixx, line 216: Now that these errors are typed, this line becomes a visible Fake/Real asymmetry: `BazaRb::Fake#transfer` validates `badge` (`'The "badge" must be a String'`), but the real `BazaRb#transfer` at `lib/baza-rb.rb:449-463` validates `recipient`, `amount`, `summary` and `job` and never looks at `badge` at all. It just does `body['badge'] = badge unless badge.nil?` on line 469.

So `baza.transfer('jeff', 1.0, 'pay', badge: 42)` raises `BazaRb::ValidationError` against the Fake and gets silently serialised into the POST body against the real client. Previously both were untyped `RuntimeError`s and the divergence was easy to miss; the typed hierarchy is what makes it worth fixing.

Out of scope for a pure re-typing PR, so I would not block on it — but it is precisely the kind of thing the contract tests in #390 are meant to catch, and #390's `test_both_reject_same_*` set does not currently cover `badge`. Either add the same guard to the real `transfer`, or drop it from the Fake, or note it on #390 so a fourth contract test covers it.

lib/baza-rb/fake.rb
```diff
@@ -262,23 +264,23 @@ def csrf
...
-    raise(RuntimeError, "The owner #{owner.inspect} is not valid") unless owner.match?(/\A.+\z/)
+    raise(BazaRb::ValidationError, 'The "owner" of the lock is nil') if owner.nil?
+    raise(BazaRb::ValidationError, 'The "owner" of the lock may not be empty') if owner.empty?
+    raise(BazaRb::ValidationError, "The owner #{owner.inspect} is not valid") unless owner.match?(/\A.+\z/)
   end
 
   def checkfile(file)
-    raise(RuntimeError, 'The file must exist') unless File.exist?(file)
-    raise(RuntimeError, 'The file must be non-empty') unless File.size(file).positive?
+    raise(BazaRb::ValidationError, 'The file must exist') unless File.exist?(file)
```
Favixx, line 283: This is the Fake side of the divergence I flagged at `lib/baza-rb.rb:324` and `:357`. `checkfile` raises `BazaRb::ValidationError` for a missing file; the real `durable_place` and `durable_save` raise the bare `BazaRb::Error` for the same condition.

Whichever direction you settle on, both sides need to move together — the value of `BazaRb::Fake` rests entirely on it failing the same way the real client fails, and this PR is the first change that can make them differ by class rather than only by message.

My recommendation is `ValidationError` on both, since an absent path is bad input. That also matches `valfile(file, must_exist: true)` in #390's `lib/baza-rb/validation.rb`, so it is the value this code converges on anyway.

While here: the message on the next line, `'The file must be non-empty'`, has no counterpart in the real client at all — `BazaRb#durable_place` only checks `File.exist?` and an upper size bound of 1024 bytes. That is a second (pre-existing) divergence in the same helper, worth a note on #390 rather than a fix here.

test/test_baza_edge.rb
```diff
@@ -23,6 +23,14 @@
 # Copyright:: Copyright (c) 2024-2026 Yegor Bugayenko
 # License:: MIT
 class TestBazaRbEdge < Minitest::Test
+  def test_error_hierarchy
+    assert_operator(BazaRb::ValidationError, :<, BazaRb::Error)
```
Favixx: These five assertions cannot fail as long as the code compiles. `assert_operator(BazaRb::ValidationError, :<, BazaRb::Error)` restates `class ValidationError < Error; end` from `lib/baza-rb.rb:45`, which is the literal declaration three lines below the class it is being compared against. The test is a mirror of the source, not a check on it; if someone deletes the hierarchy the test disappears along with it, and if someone changes a parent the test is edited in the same commit.

Also, the one relation the PR body explicitly claims to have verified — `BazaRb::Error < StandardError` — is the one relation not asserted here.

The properties actually worth locking are behavioural and non-tautological:

```ruby
def test_all_errors_rescuable_as_one
  assert_raises(BazaRb::Error) { fake_baza.push(nil, 'x', []) }
  assert_raises(BazaRb::Error) { fake_baza.pull(42.5) }
end

def test_validation_error_is_not_runtime_error
  refute_operator(BazaRb::ValidationError, :<, RuntimeError)
end
```

The second one is the regression guard that matters: it pins the deliberate incompatibility discussed on line 45, so a future "let's restore `< RuntimeError` for compatibility" has to be an explicit decision rather than a silent drift.

test/test_baza_edge.rb
```diff
@@ -265,7 +283,7 @@ def test_durable_load_rejects_range_without_hyphen
         .with(headers: { 'Range' => 'bytes=0-' })
         .to_return(status: 206, body: 'x', headers: { 'Content-Range' => 'bytes 0/10' })
       assert_includes(
-        assert_raises(RuntimeError) do
+        assert_raises(BazaRb::Error) do
```
Favixx: Asserting the base class here (and at 301, 316, 521, 537) makes these five tests weaker than they need to be. `assert_raises(BazaRb::Error)` passes for *any* error the library defines, so it would still be green if `crange` started raising `ValidationError`, `ServerFailure`, or `BadCompression` — none of which would be correct for "the server sent a malformed Content-Range".

The `assert_includes(..., 'Range is not valid ("0")')` on the message is doing all the real discrimination, which is the message-coupling issue #362 wanted to eliminate from callers; the test suite should not model the behaviour it is asking users to stop relying on.

If you adopt the `ProtocolError < Error` suggestion from `lib/baza-rb.rb:835`/`:850`, these become `assert_raises(BazaRb::ProtocolError)` and the assertion has teeth. If you keep the bare `Error`, then at minimum add one negative assertion somewhere — e.g. `refute_kind_of(BazaRb::ValidationError, err)` — so a future refactor cannot quietly reclassify a server fault as a caller fault.

test/test_fake.rb
```diff
@@ -312,18 +316,26 @@ def test_enter_rejects_multiline_name_and_badge
     baza = BazaRb::Fake.new
     assert_equal(
       'The name "test-job\nBAD" is not valid',
-      assert_raises(RuntimeError) { baza.enter("test-job\nBAD", 'test-badge', 'test-reason', 42) { 'ignored' } }.message
+      assert_raises(BazaRb::ValidationError) do
```
Favixx, line 319: All 20-odd edits in this file are a mechanical `RuntimeError` -> `BazaRb::ValidationError` swap, which is correct as far as it goes, but nothing added here or in `test_baza_edge.rb` verifies the property this PR actually puts at risk: that `BazaRb` and `BazaRb::Fake` raise the *same class* for the same invalid input.

That matters concretely — the `durable_place`/`durable_save` mismatch I flagged at `lib/baza-rb.rb:324` and `lib/baza-rb/fake.rb:283` would have been caught by a single assertion of that shape, and it went unnoticed through a green CI run on four Ruby/OS combinations. Before this PR the property held for free (everything was `RuntimeError`); this PR is what makes it breakable, so this PR should be the one that pins it.

Something as small as:

```ruby
def test_fake_and_real_agree_on_error_class
  [
    ->(b) { b.push('INVALID', 'd', []) },
    ->(b) { b.durable_place('proj', '/no/such/file') },
    ->(b) { b.transfer(nil, 1.0, 'pay') }
  ].each do |call|
    real = assert_raises(StandardError) { call.call(fake_baza) }
    fake = assert_raises(StandardError) { call.call(BazaRb::Fake.new) }
    assert_equal(real.class, fake.class)
  end
end
```

This is a strict subset of the `test_both_reject_same_*` tests in your #390, so it will be superseded rather than duplicated on rebase — but it belongs in the PR that introduces the risk, not the one after.

PR #390, opened 2026-06-05
#368/#369: shared Validation module + contract test

VasilevNStas:
## Problem

BazaRb and BazaRb::Fake had duplicated validation logic with subtle differences. 8+ issues were closed due to Fake/Real validation divergence. There was no single source of truth for validation rules.

## Solution

Created `lib/baza-rb/validation.rb` with a shared `BazaRb::Validation` module containing:
- `valname` — validates pname (nil, empty, format `/a-z0-9-/`, max 32 chars)
- `valid` — validates job/durable ID (nil, Integer, positive)
- `valfile` — validates file path (nil, empty, existence)
- `valowner` — validates owner (nil, empty, multiline)

Both `BazaRb` and `BazaRb::Fake` include the module and use its methods, ensuring consistent validation everywhere. The fake's old private helpers (`checkname`, `checkowner`, `checkfile`) were removed.

Added 3 contract tests (`test_both_reject_same_*`) that verify both implementations raise identical errors for the same invalid inputs.

## Verification
- [x] 48 edge tests, 40 fake tests — all pass
- [x] New contract tests confirm Fake and Real reject the same invalid inputs for pname, ID, and owner

Closes #368. Closes #369.


GHX5T-SOL (approved):
Approved. The current head keeps the validation refactor narrow enough to review: `BazaRb::Validation` centralizes the shared name/id/owner/file guards, both the real client and `BazaRb::Fake` include it, and the added contract tests pin fake/real parity for pname, job ID, and lock owner failures. I also checked the old fake-only zero-byte file rejection while reviewing `valfile`; the real client already accepts existing zero-byte durable files, so removing that fake-only size check reduces drift instead of creating a mismatch.

Validation I ran on exact head `77869ec99b1e0cba79a0e39810dcd4cdf07651ba`:
- `ruby -c lib/baza-rb.rb`, `ruby -c lib/baza-rb/fake.rb`, `ruby -c lib/baza-rb/validation.rb` -> Syntax OK
- focused new contract tests with coverage disabled (`/test_both_reject_same/`) -> 3 tests, 87 assertions, 0 failures
- focused fake validation/durable subset with coverage disabled -> 19 tests, 23 assertions, 0 failures
- full `test/test_baza_edge.rb` with coverage disabled -> 51 tests, 224 assertions, 0 failures
- `bundle exec rubocop` -> 13 files inspected, no offenses
- `bundle exec rake yard` -> 96.55% documented

Hosted readback: all non-rake checks pass. The Ubuntu `rake` log reaches `125 tests, 315 assertions` with one failure only in live `TestBazaLive#test_live_full_cycle` (`Job #76628 ... did not finish in 0s`); the new contract tests and the touched non-live suites pass before that. I did not run local full `bundle exec rake` because it invokes live Zerocracy API tests, which I kept out of scope.

lib/baza-rb/validation.rb
```diff
@@ -0,0 +1,63 @@
...
+      raise(RuntimeError, "The #{context} #{name.inspect} is not valid") unless name.match?(/\A[a-z0-9-]+\z/)
+      raise(RuntimeError, "The #{context} #{name.inspect} is too long") if name.length > 32
+    end
+
+    # Validate a job/durable ID.
+    #
+    # @param [Integer, nil] id The ID to validate
+    # @raise [RuntimeError] If validation fails
+    def valid(id)
+      raise(RuntimeError, 'The ID of the job is nil') if id.nil?
```
morphqdd: `valid` hardcodes "job" in every message and, unlike `valname`/`valfile`/`valowner`, takes no `context:`. That reintroduces the exact divergence this PR removes: in `fake.rb` the durable methods now call `valid(id)` and raise `The ID of the job is nil`, while the real `durable_save`/`durable_load`/`durable_lock`/`durable_unlock` (untouched by this PR) still raise `The ID of the durable is nil`. And it's uncaught, because `test_both_reject_same_id_validation` only covers `pull/finished?/stdout/exit_code/verified`. Please add `context: 'job'` and pass `'durable'` from the durable call sites.

lib/baza-rb/validation.rb
```diff
@@ -0,0 +1,63 @@
...
+    # Validate a file path.
+    #
+    # @param [String, nil] file The file path to validate
+    # @param [Boolean] must_exist Whether to check that the file exists on disk
+    # @param [String] context Human-readable field name for error messages
+    # @raise [RuntimeError] If validation fails
+    def valfile(file, must_exist: false, context: 'file')
+      raise(RuntimeError, "The \"#{context}\" is nil") if file.nil?
+      raise(RuntimeError, "The \"#{context}\" may not be empty") if file.empty?
+      raise(RuntimeError, "The file '#{file}' is absent") if must_exist && !File.exist?(file)
```
morphqdd: `valfile` is weaker than the fake's old `checkfile`, which also rejected empty files via `File.size(file).positive?` (`The file must be non-empty`). That check is gone, so `Fake#durable_place`/`durable_save` now silently accept a zero-byte file. Either restore the size check here or justify dropping it.

lib/baza-rb.rb
```diff
@@ -35,6 +36,8 @@
 class BazaRb
   DEFAULT_CHUNK_SIZE = 1_000_000
 
+  include Validation
```
morphqdd, line 41: The module is only applied to a subset of the real class: `push`'s data/meta checks, all `durable_*`, `transfer`, `fee`, and `enter` still validate inline with `raise(RuntimeError, ...)`. So the "single source of truth" is only half-established, and the fake/real divergence persists (and is newly created — see the `valid` comment) for exactly the methods left inline. Either convert them in this PR too, or narrow #368/#369's scope to what is actually unified.

lib/baza-rb.rb
```diff
@@ -107,10 +110,7 @@ def balance
   # @param [Integer] chunk_size Maximum size of one chunk
   # @raise [ServerFailure] If the push operation fails
   def push(pname, data, meta, chunk_size: DEFAULT_CHUNK_SIZE)
-    raise(RuntimeError, 'The "name" of the job is nil') if pname.nil?
-    raise(RuntimeError, 'The "name" of the job may not be empty') if pname.empty?
-    raise(RuntimeError, "The name #{pname.inspect} is not valid") unless pname.match?(/\A[a-z0-9-]+\z/)
-    raise(RuntimeError, "The name #{pname.inspect} is too long") if pname.length > 32
+    valname(pname, context: 'name')
```
morphqdd, line 118: Message regression: on `master`, real `push(nil, ...)` raised `The "name" of the job is nil`; via `valname(pname, context: 'name')` it now raises `The "name" is nil` — the "of the job" phrasing is lost, and the `context:` mechanism can't reproduce it. If any callers/tests assert exact message strings, this is a breaking change.

lib/baza-rb/fake.rb
```diff
@@ -258,24 +260,9 @@ def csrf
...
-  def checkowner(owner)
-    raise(RuntimeError, 'The "owner" of the lock is nil') if owner.nil?
-    raise(RuntimeError, 'The "owner" of the lock may not be empty') if owner.empty?
-    raise(RuntimeError, "The owner #{owner.inspect} is not valid") unless owner.match?(/\A.+\z/)
-  end
-
-  def checkfile(file)
-    raise(RuntimeError, 'The file must exist') unless File.exist?(file)
-    raise(RuntimeError, 'The file must be non-empty') unless File.size(file).positive?
+  # Do nothing, here for signature parity.
```
morphqdd: This leftover `checkid` is dead code — every call site now uses `valid(id)` — and the comment `Do nothing, here for signature parity` is factually wrong (the method raises). Its messages (`The ID must be an Integer`) also differ from `valid`'s. Please remove it.

test/test_baza_edge.rb
```diff
@@ -979,6 +981,48 @@ def test_upload_raises_when_rewinds_exceed_retries
     end
   end
```
morphqdd, line 1093: Great addition — contract tests are exactly the right tool here. But they only cover `push`, the job-ID methods, and `lock/unlock`. They miss the `durable_*` methods and file validation, which is precisely where the remaining divergence lives (see the `valid` and `valfile` comments). Extending them there would have caught the durable ID-message mismatch.

lib/baza-rb/validation.rb
```diff
@@ -0,0 +1,63 @@
+# frozen_string_literal: true
+
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 Zerocracy
```
morphqdd, line 3: Copyright holder is inconsistent within this new file: the SPDX header says `Copyright (c) 2024-2026 Zerocracy` while the `Copyright::` line below says `Yegor Bugayenko`. The rest of the repo uses `Yegor Bugayenko` — please align.

lib/baza-rb/validation.rb
```diff
@@ -0,0 +1,63 @@
...
+  module Validation
+    # Validate a product/job/durable name (pname).
+    #
+    # @param [String, nil] name The name to validate
+    # @param [String] context Human-readable field name for error messages
+    # @raise [RuntimeError] If validation fails
+    def valname(name, context: 'name')
+      raise(RuntimeError, "The \"#{context}\" is nil") if name.nil?
+      raise(RuntimeError, "The \"#{context}\" may not be empty") if name.empty?
+      raise(RuntimeError, "The #{context} #{name.inspect} is not valid") unless name.match?(/\A[a-z0-9-]+\z/)
```
morphqdd: Inconsistent message shape: nil/empty use a quoted `"#{context}"`, but the format/length messages use a bare `#{context}`. With `context: 'pname of the product'` (from `lock`), this renders `The pname of the product "x" is not valid`, which reads awkwardly and differs from the previous `The name "x" is not valid`. Consider a consistent wording across all four raises.

lib/baza-rb/validation.rb
```diff
@@ -0,0 +1,63 @@
+# frozen_string_literal: true
+
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026 Zerocracy
+# SPDX-License-Identifier: MIT
+
+# Shared validation methods for BazaRb and BazaRb::Fake.
+#
+# Every method here raises RuntimeError on invalid input.
```
morphqdd: Heads-up on a collision with the sibling PR #389, which moves every error to `BazaRb::Error`. This module hardcodes `RuntimeError` in all four helpers; whichever lands second will need to update the other. Worth coordinating so the shared module raises the new base class from the start.

lib/baza-rb.rb
```diff
@@ -37,6 +38,10 @@ class BazaRb
   DEFAULT_CHUNK_SIZE = 1_000_000
 
   include Compress
+  include Validation
+
+  # When an invalid argument is provided.
+  class ValidationError < StandardError; end
```
Favixx: This is the line with the widest blast radius in the PR. Every `raise(RuntimeError, ...)` in this file became `raise(BazaRb::ValidationError, ...)`, and `ValidationError` descends from `StandardError`, not from `RuntimeError` — so `rescue RuntimeError` around a `BazaRb` call no longer catches what it used to. `rescue => e` still works, which is what makes this quiet.

To be fair to the change first: I searched the whole organisation for `rescue RuntimeError` and it returns zero hits, so nothing inside `zerocracy` breaks the day this merges. The problem is one step further out.

`fbe` depends on this gem (`fbe.gemspec`: `add_dependency('baza.rb', '~>0.9')`), and `fbe/lib/fbe/enter.rb` calls `baza.enter(...)` — one of the methods this PR converts. Line 20 of that file is a published YARD contract:

> `@raise [RuntimeError] If badge, why, or required globals are nil`

Those nil checks are `BazaRb#enter`'s. After this PR they raise something that is not a `RuntimeError`, so `fbe` ships a documented promise that this change falsifies — and `fbe` is a public gem whose users are not in this organisation and cannot be grepped.

`class ValidationError < RuntimeError; end` gives callers the specific type they want, keeps every existing rescue working, and leaves `fbe`'s documentation true. If descending from `StandardError` is deliberate, then this is a breaking change: it needs to be in the description, in the version bump, and coordinated with `fbe` — which pins `~>0.9` while `judges-action` pins `~>0.11`, so the two consumers will not pick it up at the same time.

lib/baza-rb.rb
```diff
@@ -269,10 +253,10 @@ def unlock(pname, owner)
   # @return [Integer] The ID of the job on the server
   # @raise [ServerFailure] If the job doesn't exist or retrieval fails
   def recent(name)
-    raise(RuntimeError, 'The "name" of the job is nil') if name.nil?
-    raise(RuntimeError, 'The "name" of the job may not be empty') if name.empty?
-    raise(RuntimeError, "The name #{name.inspect} is not valid") unless name.match?(/\A[a-z0-9-]+\z/)
-    raise(RuntimeError, "The name #{name.inspect} is too long") if name.length > 32
+    raise(BazaRb::ValidationError, 'The "name" of the job is nil') if name.nil?
```
Favixx: `recent` still hand-writes the four checks that `valname` was extracted to hold — and `BazaRb::Fake#recent` was converted, so the two implementations now disagree about the message for the same input.

`BazaRb#recent(nil)` → `The "name" of the job is nil`
`BazaRb::Fake#recent(nil)` → `The "name" is nil`

That is a fresh instance of the divergence this PR exists to eliminate, created by converting one side and not the other. `valname(name)` here makes them agree, and deletes four lines.

The same applies to `name_exists?` on line 273, whose fake counterpart is also already on `valname`.

lib/baza-rb.rb
```diff
@@ -286,10 +270,10 @@ def recent(name)
   # @param [String] pname The name of the product on the server
   # @return [Boolean] TRUE if such name exists
   def name_exists?(pname)
-    raise(RuntimeError, 'The "pname" of the product is nil') if pname.nil?
-    raise(RuntimeError, 'The "pname" of the product may not be empty') if pname.empty?
-    raise(RuntimeError, "The name #{pname.inspect} is not valid") unless pname.match?(/\A[a-z0-9-]+\z/)
-    raise(RuntimeError, "The name #{pname.inspect} is too long") if pname.length > 32
+    raise(BazaRb::ValidationError, 'The "pname" of the product is nil') if pname.nil?
```
Favixx: Second of the two methods left behind — `name_exists?` here versus `valname(pname)` in the fake.

Worth noting what these two survivors have in common with the ones that were converted: nothing. `pull`, `finished?`, `stdout`, `exit_code`, `verified`, `lock`, `unlock` all moved; `recent`, `name_exists?`, and the non-name arguments of `push`, `fee` and `enter` did not. If the rule was "convert what maps cleanly onto an existing validator", then these two do map cleanly and were simply missed.

A quick way to keep it from happening again: once every call site is converted, `grep -c 'raise(BazaRb::ValidationError' lib/baza-rb.rb` should be small and each survivor should be a genuinely one-off rule. Today it is 20-odd.

lib/baza-rb.rb
```diff
@@ -110,14 +115,11 @@ def balance
...
-    raise(RuntimeError, 'The "name" of the job is nil') if pname.nil?
-    raise(RuntimeError, 'The "name" of the job may not be empty') if pname.empty?
-    raise(RuntimeError, "The name #{pname.inspect} is not valid") unless pname.match?(/\A[a-z0-9-]+\z/)
-    raise(RuntimeError, "The name #{pname.inspect} is too long") if pname.length > 32
-    raise(RuntimeError, 'The "data" of the job is nil') if data.nil?
-    raise(RuntimeError, 'The "data" of the job may not be empty') if data.empty?
-    raise(RuntimeError, 'The "meta" of the job is nil') if meta.nil?
-    raise(RuntimeError, 'The "meta" of the job must be an Array') unless meta.is_a?(Array)
+    valname(pname, context: 'name')
+    raise(BazaRb::ValidationError, 'The "data" of the job is nil') if data.nil?
```
Favixx: `push` is the method the PR description leads with, and its non-name arguments still diverge from the fake:

| | real | fake |
|---|---|---|
| empty data | `The "data" of the job may not be empty` | `The data must be non-empty` |
| bad meta | `The "meta" of the job must be an Array` | `The meta must be an array` |

The fake also folds nil-meta into the `is_a?(Array)` check while this one raises a separate `The "meta" of the job is nil` first, so `push('x', 'y', nil)` produces two different messages depending on which implementation you hold.

The module has no validator for either shape, which is the real gap: `valdata` and an `valarray`-ish check would finish the job, and the three contract tests would then have something to assert about `push` rather than only about pname, ID and owner.

lib/baza-rb.rb
```diff
@@ -230,12 +222,8 @@ def verified(id)
   # @raise [RuntimeError] If the name is already locked
   # @raise [ServerFailure] If the lock operation fails
   def lock(pname, owner)
-    raise(RuntimeError, 'The "pname" of the product is nil') if pname.nil?
-    raise(RuntimeError, 'The "pname" of the product may not be empty') if pname.empty?
-    raise(RuntimeError, "The name #{pname.inspect} is not valid") unless pname.match?(/\A[a-z0-9-]+\z/)
-    raise(RuntimeError, "The name #{pname.inspect} is too long") if pname.length > 32
-    raise(RuntimeError, 'The "owner" of the lock is nil') if owner.nil?
-    raise(RuntimeError, 'The "owner" of the lock may not be empty') if owner.empty?
+    valname(pname, context: 'pname of the product')
```
Favixx: The `context:` interpolation puts the quotes around whatever string it is given, so this produces `The "pname of the product" is nil` — quoting a phrase rather than a field name. The message it replaces was `The "pname" of the product is nil`, which reads correctly.

Same on line 242 with `'pname of the job'`.

Two ways out. Either pass `context: 'pname'` and accept a slightly shorter message, or give `valname` a second keyword for the trailing qualifier so the template becomes `The "#{context}" of the #{owner} is nil`. The first is what the other twelve call sites already do.

lib/baza-rb.rb
```diff
@@ -517,14 +485,14 @@ def fee(tab, amount, summary, job)
...
   def enter(pname, badge, why, job)
-    raise(RuntimeError, 'The "pname" is nil') if pname.nil?
-    raise(RuntimeError, 'The "pname" may not be empty') if pname.empty?
-    raise(RuntimeError, 'The "badge" is nil') if badge.nil?
-    raise(RuntimeError, 'The "badge" may not be empty') if badge.empty?
-    raise(RuntimeError, 'The "why" is nil') if why.nil?
-    raise(RuntimeError, 'The "why" may not be empty') if why.empty?
-    raise(RuntimeError, 'The "job" must be an Integer') unless job.nil? || job.is_a?(Integer)
-    raise(RuntimeError, 'The "job" must be positive') unless job.nil? || job.positive?
+    raise(BazaRb::ValidationError, 'The "pname" is nil') if pname.nil?
```
Favixx, line 476: `enter` checks `pname` for nil and empty only, while `push`, `lock`, `unlock` and `name_exists?` all run the full `valname` on the same concept — so a product name with uppercase letters, or forty characters long, is rejected by four methods and accepted by this one.

That is a divergence inside a single class, which the shared module cannot help with because nobody called it here. If `enter` deliberately takes a looser name than the rest, it is worth a comment saying so; if not, `valname(pname)` is the line.

Worth extra care on this method specifically: it is the one `BazaRb` entry point with a known external caller. `fbe/lib/fbe/enter.rb` wraps it, passes `options.job_name` straight through as `pname`, and documents the resulting failure as a `RuntimeError`. Tightening the rule here changes what `fbe` accepts; changing the exception class changes what `fbe` documents. Both are fine to do, but they should be done knowingly rather than as a side effect.

lib/baza-rb.rb
```diff
@@ -441,19 +406,22 @@ def durable_find(pname, file)
...
-      raise(RuntimeError, 'The ID must be an Integer') unless job.is_a?(Integer)
-      raise(RuntimeError, 'The ID must be positive') unless job.positive?
+    raise(BazaRb::ValidationError, 'The "amount" is nil') if amount.nil?
+    unless amount.is_a?(Float) || amount.is_a?(BigDecimal)
+      raise(BazaRb::ValidationError, 'The "amount" must be Float or BigDecimal')
     end
+    raise(BazaRb::ValidationError, 'The "amount" must be positive') unless amount.positive?
+    raise(BazaRb::ValidationError, 'The "summary" is nil') if summary.nil?
+    raise(BazaRb::ValidationError, "The summary #{summary.inspect} is empty") if summary.empty?
+    valid(job) unless job.nil?
```
Favixx: `valid(job) unless job.nil?` — the optional-argument idiom, repeated here and at `fake.rb:217` and `fake.rb:255`. Three copies of a guard that the validator could own: `valid(job, optional: true)`, or a `valid?`-style wrapper.

Minor on its own. It is worth mentioning because it is the same pressure that produced the duplication this PR is removing: a rule that the module does not express gets written at the call site, and then written again slightly differently somewhere else.

lib/baza-rb/fake.rb
```diff
@@ -37,10 +39,10 @@ def whoami
   # @param [Integer] chunk_size Size of each chunk in bytes (accepted for signature parity with BazaRb#push)
   # @return [Integer] Always returns 42 as the fake job ID
   def push(name, data, meta, chunk_size: BazaRb::DEFAULT_CHUNK_SIZE) # rubocop:disable Lint/UnusedMethodArgument
-    checkname(name)
-    raise(RuntimeError, 'The "data" of the job is nil') if data.nil?
-    raise(RuntimeError, 'The data must be non-empty') if data.empty?
-    raise(RuntimeError, 'The meta must be an array') unless meta.is_a?(Array)
+    valname(name)
+    raise(BazaRb::ValidationError, 'The "data" of the job is nil') if data.nil?
+    raise(BazaRb::ValidationError, 'The data must be non-empty') if data.empty?
```
Favixx: These two messages are the fake's half of the `push` divergence described on `baza-rb.rb:119` — `The data must be non-empty` against the real `The "data" of the job may not be empty`, and `The meta must be an array` against `The "meta" of the job must be an Array`.

Worth fixing in this PR rather than a follow-up, because the PR's own description says the module ensures "consistent validation everywhere", and `push` is the first example it gives. Right now `push` is the method where the claim is least true.

lib/baza-rb/fake.rb
```diff
@@ -94,25 +96,25 @@ def verified(id)
...
+    valowner(owner)
   end
 
   # Get the ID of the job by the name.
   #
   # @param [String] name The name of the job on the server
   # @return [Integer] The ID of the job on the server
   def recent(name)
-    checkname(name)
+    valname(name)
```
Favixx, line 116: This is the converted side of the pair — `valname(name)` here, four hand-written raises in `BazaRb#recent`. Nothing is wrong with this line; it is the evidence that the other side was missed.

If you want a cheap guard against the general case, the contract tests could iterate over the methods both classes expose rather than testing three fixed inputs: for each shared method name, feed the same invalid argument to both and assert the messages match. That turns "we checked pname, ID and owner" into "we checked everything they both implement", and it would have failed on `recent` today.

lib/baza-rb/validation.rb
```diff
@@ -0,0 +1,63 @@
...
+      end
+      raise(BazaRb::ValidationError, "The \"#{context}\" #{name.inspect} is too long") if name.length > 32
+    end
+
+    # Validate a job/durable ID.
+    #
+    # @param [Integer, nil] id The ID to validate
+    # @param [String] context Human-readable noun for error messages (default 'job')
+    # @raise [BazaRb::ValidationError] If validation fails
+    def valid(id, context: 'job')
```
Favixx: `valid` is the one name here that fights its meaning. Every other method is `val` plus the thing it checks — `valname`, `valfile`, `valowner` — and this one is `val` plus `id`, which collides with the ordinary English adjective. `valid(id)` at a call site reads as a predicate returning a boolean, and it raises instead.

`valjob` matches the default `context:`, or `valnum` matches the shape it actually checks (a positive Integer, which is also what `durable` IDs are). Either survives a reader who has not seen this file.

lib/baza-rb/validation.rb
```diff
@@ -0,0 +1,63 @@
...
+# guarantee they reject the same set of invalid inputs.
+class BazaRb
+  # Validation helpers.
+  module Validation
+    # Validate a product/job/durable name (pname).
+    #
+    # @param [String, nil] name The name to validate
+    # @param [String] context Human-readable field name for error messages
+    # @raise [BazaRb::ValidationError] If validation fails
+    def valname(name, context: 'name')
```
Favixx, line 20: The header of this file promises that "Every method here raises BazaRb::ValidationError on invalid input", and for a non-String that is not true: `valname(42)` reaches `42.empty?` and raises `NoMethodError`, which is not a `ValidationError` and, after the change on `baza-rb.rb:44`, is not caught by anything rescuing either.

`valid` guards its type explicitly with `is_a?(Integer)`, so the module is inconsistent with itself as well. A `unless name.is_a?(String)` line here, and the same in `valfile` and `valowner`, closes it and makes the header true.

lib/baza-rb/validation.rb
```diff
@@ -0,0 +1,63 @@
...
+      raise(BazaRb::ValidationError, "The \"#{context}\" must be positive") unless id.positive?
+    end
+
+    # Validate a file path.
+    #
+    # @param [String, nil] file The file path to validate
+    # @param [Boolean] must_exist Whether to check that the file exists on disk
+    # @param [String] context Human-readable field name for error messages
+    # @raise [RuntimeError] If validation fails
+    def valfile(file, must_exist: false, context: 'file')
```
Favixx, line 67: `must_exist` does two things: it checks existence and it checks that the file is non-empty. A caller reading `valfile(file, must_exist: true)` at `baza-rb.rb:298` has no way to guess the second rule, and a caller who wants "must exist, may be empty" has no way to ask for it.

Splitting into `must_exist:` and `must_have_content:` says what happens; if the pair is always wanted together, then naming the single flag `must_have_data:` at least names the stronger of the two conditions.

lib/baza-rb/validation.rb
```diff
@@ -0,0 +1,63 @@
...
+      raise(BazaRb::ValidationError, "The \"#{context}\" is nil") if file.nil?
+      raise(BazaRb::ValidationError, "The \"#{context}\" may not be empty") if file.empty?
+      raise(BazaRb::ValidationError, "The file '#{file}' is absent") if must_exist && !File.exist?(file)
```
Favixx, line 71: `File.exist?` then `File.empty?` are two separate syscalls on the same path, so between them the file can be deleted — and `File.empty?` on a path that has just disappeared raises `Errno::ENOENT`, not a `ValidationError`.

Narrow window and an unusual failure, but this gem's own `#385` is currently open against a TOCTOU race in `durable_place`, so it is evidently a class of bug this codebase cares about. `File.size?` returns nil for both "absent" and "empty" in one call, which collapses the two checks into one and removes the window.

lib/baza-rb/validation.rb
```diff
@@ -0,0 +1,63 @@
...
+
+    # Validate an owner string.
+    #
+    # @param [String, nil] owner The owner to validate
+    # @param [String] context Human-readable field name for error messages
+    # @raise [RuntimeError] If validation fails
+    def valowner(owner, context: 'owner')
+      raise(BazaRb::ValidationError, "The \"#{context}\" is nil") if owner.nil?
+      raise(BazaRb::ValidationError, "The \"#{context}\" may not be empty") if owner.empty?
+      raise(BazaRb::ValidationError, "The \"#{context}\" #{owner.inspect} is not valid") unless owner.match?(/\A.+\z/)
```
Favixx, line 84: `match?(/\A.+\z/)` is an obscure way to say "one line only": it works because `.` does not cross a newline and `\z` is the absolute end, so `"a\nb"` and `"a\n"` both fail. A reader who does not hold both of those facts at once will read this as "not empty", which the line above already checked.

`owner.include?("\n")` says it directly, and the message can then say what is actually wrong — `The "owner" may not span several lines` — instead of the generic `is not valid`.

lib/baza-rb/validation.rb
```diff
@@ -0,0 +1,63 @@
...
+      raise(BazaRb::ValidationError, "The \"#{context}\" must be an Integer") unless id.is_a?(Integer)
+      raise(BazaRb::ValidationError, "The \"#{context}\" must be positive") unless id.positive?
+    end
+
+    # Validate a file path.
+    #
+    # @param [String, nil] file The file path to validate
+    # @param [Boolean] must_exist Whether to check that the file exists on disk
+    # @param [String] context Human-readable field name for error messages
+    # @raise [RuntimeError] If validation fails
```
Favixx: `@raise [RuntimeError]` here and on line 56, while `valname` and `valid` correctly document `@raise [BazaRb::ValidationError]`. Two of the four doc comments describe the behaviour this PR replaced.

Worth fixing carefully rather than mechanically: if `ValidationError` ends up inheriting from `RuntimeError` as suggested on `baza-rb.rb:44`, these two become accidentally true, which is not the same as being right.

.rubocop.yml
```diff
@@ -43,6 +43,12 @@ Style/StaticClass:
 Metrics/AbcSize:
   Exclude:
     - 'lib/baza-rb.rb'
+Metrics/CyclomaticComplexity:
```
Favixx, line 46: Two new complexity exemptions for `lib/baza-rb.rb` in a PR that removes 103 lines from it and moves four validators out. Extracting duplicated guard clauses should lower cyclomatic complexity, not push it past a threshold it was previously under.

My guess is that the file was already over both limits and `Metrics/AbcSize` was the only one being tripped before — in which case these two lines are pre-existing debt surfacing rather than something this PR caused, and saying so in the description would settle it. If instead a method really did get more branchy, it is worth finding which one, because that is the opposite of what the change is for.

Either way, suppressing a metric in the same commit that claims to simplify is the kind of thing that is much easier to explain now than to explain to whoever reads `git blame` in a year.

PR #406, opened 2026-06-26
Validate non-empty summary in fee()

edmoffo:
The fee() method in lib/baza-rb.rb guarded against a nil summary but accepted an empty string, so baza.fee('unknown', 1.0, '', 42) reached the server with an empty summary. The transfer() method already rejects both nil and empty on the line above. Mirror that on the fee() method and add an edge test in test/test_baza_edge.rb following the same pattern as test_transfer_raises_when_summary_is_empty.

Closes #401.

---
_Generated by [Claude Code](https://claude.ai/code)_

### DIFF
--- a/lib/baza-rb.rb
+++ b/lib/baza-rb.rb
@@ -458,6 +458,7 @@ class BazaRb
     raise(RuntimeError, 'The "job" is nil') if job.nil?
     raise(RuntimeError, 'The "job" must be Integer') unless job.is_a?(Integer)
     raise(RuntimeError, 'The "summary" is nil') if summary.nil?
+    raise(RuntimeError, "The summary #{summary.inspect} is empty") if summary.empty?
     id = nil
     elapsed(@loog, level: Logger::INFO) do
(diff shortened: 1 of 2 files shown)


lib/baza-rb.rb
```diff
@@ -458,6 +458,7 @@ def fee(tab, amount, summary, job)
     raise(RuntimeError, 'The "job" is nil') if job.nil?
     raise(RuntimeError, 'The "job" must be Integer') unless job.is_a?(Integer)
     raise(RuntimeError, 'The "summary" is nil') if summary.nil?
+    raise(RuntimeError, "The summary #{summary.inspect} is empty") if summary.empty?
```
morphqdd, line 487: The real client is fixed, but `BazaRb::Fake#fee` still doesn't reject an empty summary — so fake and real diverge again on this input (the same class of fake/real drift #390 is trying to eliminate). Worth mirroring the guard in `fake.rb` and asserting it in a fake test too.

PR #407, opened 2026-06-26
fix(#404): send valid Accept header

terminalchai:
Fixes #404.

This changes the download request header from `Accept: *` to the valid universal media range `Accept: */*`, and adds a WebMock regression test for the header sent by `download`.

I could not run the Ruby test suite locally because Ruby/Bundler are not installed on this Windows PATH and Docker Desktop's Linux engine is not running in this environment. I did run `git diff --check`, and the patch is limited to the header change plus the regression test.


lib/baza-rb.rb
```diff
@@ -768,7 +768,7 @@ def download(uri, file)
                   uri.to_s,
                   method: :get,
                   headers: headers.merge(
-                    'Accept' => '*',
+                    'Accept' => '*/*',
```
morphqdd, line 771: Good — matches RFC 9110: `Accept` values must be `type/subtype`, so `*/*` is the correct wildcard and some stricter servers/proxies reject a bare `*`.

PR #408, opened 2026-06-30
#402: gemspec missing runtime deps on base64 and logger

VasilevNStas:
Fixes #402.

- Adds `s.add_dependency('base64', '~>0.3')` to gemspec
- Adds `s.add_dependency('logger', '~>1.6')` to gemspec  
- Extends CI matrix to include Ruby 3.4 alongside 3.3

`base64` was demoted from default to bundled gem in Ruby 3.4.
`logger` follows the same path in Ruby 3.5.
Downstream consumers on 3.4+ get `LoadError` without explicit deps.


baza.rb.gemspec
```diff
@@ -23,8 +23,10 @@ Gem::Specification.new do |s|
   s.files = `git ls-files | grep -v -E '^(test/|\\.|renovate)'`.split($RS)
   s.rdoc_options = ['--charset=UTF-8']
   s.extra_rdoc_files = ['README.md', 'LICENSE.txt']
+  s.add_dependency('base64', '~>0.3')
```
morphqdd, line 26: Adding 3.4 to the matrix is the right move here — it's what makes this dependency fix meaningful (on 3.3 the gems are still bundled, so the bug is invisible). Good.

PR #409, opened 2026-06-30
#400: fee() missing positive integer validation for job parameter

VasilevNStas:
Fixes #400.

`fee()` checked `job` for nil and Integer type but did not validate that it is positive, unlike `transfer()` which has `unless job.positive?`.

- Adds `raise(RuntimeError, 'The "job" must be positive') unless job.positive?` to `fee` in `lib/baza-rb.rb`
- Same fix in `lib/baza-rb/fake.rb`
- Adds test `test_fee_raises_when_job_is_not_positive` to both `test_baza_edge.rb` and `test_fake.rb`


lib/baza-rb.rb
```diff
@@ -480,6 +480,7 @@ def fee(tab, amount, summary, job)
     raise(RuntimeError, 'The "amount" must be positive') unless amount.positive?
     raise(RuntimeError, 'The "job" is nil') if job.nil?
     raise(RuntimeError, 'The "job" must be Integer') unless job.is_a?(Integer)
+    raise(RuntimeError, 'The "job" must be positive') unless job.positive?
```
morphqdd, line 486: Same 'mirror `transfer`' rationale, for a follow-up: `transfer` also rejects an *empty* summary (`summary.empty?`), whereas `fee` (just below) still only checks `summary.nil?`, so `fee('x', 1.0, '', 42)` passes an empty summary to the server. That specific gap is handled in #406 — flagging so the two land consistently.

